PR #417, opened 2020-10-01
Adds encapsulation to a few fields

Imbahil:
I changed some access modifiers from default to private, they had getters and mutators and were still package-private, so I changed that.


buddhika75:
Thank you so much for the improvements

PR #4352, opened 2024-03-28
Issue#4330

DARKDRAGON-LK:


DARKDRAGON-LK:
### Fixed Following

Lab Summaries should be renamed as 'Lab Analytics'

PR #4759, opened 2024-04-24
Issue#4741

buddhika75:


src/main/webapp/channel/manage_booking.xhtml
PasinduW99, line 1: In the method 'navigateBackToBookingsLoagingBillSessions()' the word Loading is misspelled.

PasinduW99 (approved):

PR #4798, opened 2024-04-25
Issue#4786

DARKDRAGON-LK:


DARKDRAGON-LK:
### Fixed following Issue : 

> Added Channel cancel bills wasn't displayed on the All Financial Transaction Summary

PR #4816, opened 2024-04-25
Issue#4791

Senula88:


src/main/java/com/divudi/bean/pharmacy/TransferIssueController.java
```diff
@@ -142,12 +142,14 @@ public void setUserStockContainer(UserStockContainer userStockContainer) {
     }
 
     public void remove(BillItem billItem) {
-        if (billItem.getTransUserStock().isRetired()) {
-            JsfUtil.addErrorMessage("This Item Already removed");
-            return;
-        }
+        if (billItem.getTransUserStock() != null) {
```
DARKDRAGON-LK, line 145: If there is a null item the code will run

DARKDRAGON-LK (changes requested):

src/main/java/com/divudi/bean/pharmacy/TransferIssueController.java
DARKDRAGON-LK: like this
```
public void remove(BillItem billItem) {
        if (billItem.getTransUserStock() == null) {
		JsfUtil.addErrorMessage("There is no Item to removed");
                return;            

            }

       if (billItem.getTransUserStock().isRetired()) {
                JsfUtil.addErrorMessage("This Item Already removed");
                return;
}
```

DARKDRAGON-LK (approved):
No Issues Found

PR #4908, opened 2024-04-29
Issue#4885

DamithDeshan:


src/main/java/com/divudi/bean/common/BillController.java
```diff
@@ -3167,6 +3168,32 @@ public void updateReferingDoctor() {
         billFacade.edit(bill);
         JsfUtil.addSuccessMessage("Ref Doctor Updated");
     }
+    
+    public String findBatchBillSessionID(String deptID){
```
buddhika75, line 3172: We can get the id to search here instead of deptID
Also I think we can get it from the object without using JPQL

Please discuss

src/main/java/com/divudi/bean/common/OpdPreBillController.java
```diff
@@ -840,6 +840,8 @@ private void saveBillItemSessions() {
             }
         }
     }
+    
+    private PreBill batchBill;
```
buddhika75, line 844: We have to take this up where class variables are defined

src/main/java/com/divudi/bean/common/OpdPreSettleController.java
```diff
@@ -291,29 +291,126 @@ private void saveSettlingBatchBill() {
         getSaleBill().setBillClassType(BillClassType.BilledBill);
         getSaleBill().setBillType(BillType.OpdBill);
         getSaleBill().setBillTypeAtomic(BillTypeAtomic.OPD_BATCH_BILL_PAYMENT_COLLECTION_AT_CASHIER);
-
         getSaleBill().setDepartment(getSessionController().getLoggedUser().getDepartment());
         getSaleBill().setInstitution(getSessionController().getLoggedUser().getDepartment().getInstitution());
-
         getBillBean().setPaymentMethodData(getSaleBill(), getSaleBill().getPaymentMethod(), paymentMethodData);
-
+        getBillBean().setBills(billsOfBatchBillPre);
```
buddhika75, line 297: here we have to assign the bills newly created with the batch bill, not those of preBill

PR #4948, opened 2024-05-02
Issue#4945

DARKDRAGON-LK:


DARKDRAGON-LK:
### Fixed Following Issue

GUI Improvement of Channeling -> Channel Reports -> Income Reports -> Income With Agent Bookings
/channel/income_with_agent_bookings.xhtml

PR #5077, opened 2024-05-09
Issue#4854

DARKDRAGON-LK:


DARKDRAGON-LK:
### Fixed Following Issue:

> Need to print original bill again as an admin option

**Notes**

- There is a New Privilage Under Channelling > Administrator > Channelling Reprint Original Bill

PR #5355, opened 2024-05-23
Issue#5354

buddhika75:
Need to set the required attributes under Application Options

Have to check all events where SMS is generated to check SMS is working.

All instances need to update Application Options related to SMS sending.


src/main/java/com/divudi/bean/common/OpdTabPreBillController.java
```diff
@@ -1353,10 +1353,11 @@ public void addToBill() {
             return;
         }
         clearBillItemValues();
-        //JsfUtil.addSuccessMessage("Item Added");
+        JsfUtil.addSuccessMessage("Item Added");
     }
 
     public void clearBillItemValues() {
+        System.out.println("clearBillItemValues");
```
damithdeshan98: Remove the "System.out.println"

DamithDeshan (commented):
Remove the "System.out.println"

DamithDeshan (approved):
Approved

PR #5447, opened 2024-05-28
Issue#5215

DARKDRAGON-LK:


DARKDRAGON-LK:
### Fixed Following Issue:
(/channel/manage_booking.xhtml)

> On-call booking can be settled using on-call(it was displayed on the channel booking page as paid on call)

PasinduW99 (approved):
Reviewed

PR #5572, opened 2024-06-02
Issue#5515

DARKDRAGON-LK:


DARKDRAGON-LK:
### Fixed Following Issue:
After Making a Payment to the Doctor , there will be a change in summary of the channeling.

cashier/cashier_summary.xhtml
(Arogya Hospital)

PR #5578, opened 2024-06-03
Issue#5577

DARKDRAGON-LK:


DARKDRAGON-LK:
### Fixed Following Issues:
cashier/shift_end_summery_bill.xhtml
Shift End Funds

> Can Receive Funds without Starting Shift in Financial Transaction Manager

PR #7545, opened 2024-09-28
Issue#7477

Pubudu-Piyankara:
When cc bill canceled the payment amount should be out from drawer. It is done.


SandithP2001 (approved):
Reviewed

PR #7791, opened 2024-10-06
Issue#7781

Pubudu-Piyankara:
Cannot Safe Cash Deposit more than the Cash Balance in the Drawer
Path: Payment/Handover Management/Deposit To Safe/Bank

The Cash Balance of Drawer never be Minus


SandithP2001 (approved):
Reviewed

PR #7827, opened 2024-10-07
Issue#7804

buddhika75:
Need to pull latest to the staging and then check
Now under OPD > Billing > Collecting Centre Billing>
Can select a CC without a Null Pointer Error
Also can select items without errors


Pubudu-Piyankara (approved):
reviewed

PR #7884, opened 2024-10-09
Issue##7721

SandithP2001:
Ask the doctor and complete it.

1. Removed "Search" Button. - It is not working correctly and does not need 2 search buttons
2.  Renamed "Search All" button as "Search"
3. Add a paginator to the result table

Now when clicking on the Search button it will display only due amounts....Not display refunded or canceled bills.


Pubudu-Piyankara (approved):

DamithDeshan (approved):
No Issues

PR #7998, opened 2024-10-14
Issue#7951

SandithP2001:
From Date
To Date
CC name 
added before to the table as requested

PHN column name renamed as MRN as requested

Department - Cannot search using the department ----> this requested but there is no department field to search???


DamithDeshan (approved):
No Issues

Issue #8062, opened 2024-10-17
Patient Deposit Returns are not correctly Reflicted in reports


buddhika75:
When a patient deposit is made, all reports are correctly updated as in cash in

But when patient deposits are returned, the reports are not updated as cash out. Instead it adds to cash in

TO test it, go to patient deposit, make a deposit. whck with daily return. then make a return, then check daily return. you will see an increase in cash in even after returning.

buddhika75:
After creating a PatientPaymentRefundBill, the total comes correctly as a minus, but the net total is plus.

buddhika75:
Now if you make a patient deposit return, the values are reflected correctly as cash out in reports.

PR #8063, opened 2024-10-17
Issue#7837

buddhika75:
Issues corrected
1. THose are filtered only by date, now filter to date and time
2, When more than one is given for quantities to items like dressings, they are counted under the qty in OPD collection
3. If a patient deposit is returned, it will apprear as minus, but that needs to be corrected at the time of retursning. an issues was reaised for that - [#8062](https://github.com/hmislk/hmis/issues/8062) #8062


SandithP2001 (approved):
Reviewed

PR #8064, opened 2024-10-17
Issue#8062

buddhika75:
Now if you make a patient deposit return, the values are reflected correctly as cash out in reports.


SandithP2001 (approved):
Reviewed

PR #8100, opened 2024-10-19
Issue#7989

Pubudu-Piyankara:
reports/cashier_reports/cashier_detailed.xhtml
The refund amount wasn't reduced from net amount
Cashier Report/Cashier Details

Agency Management
Agency Accept Deposit isn't updated

Patient Deposit Management
Return Accept patient Deposit isn't reduced from total


PasinduW99 (approved):
Done

PR #8260, opened 2024-10-25
Issue#8242

SandithP2001:
Slip- The date was missing ----> FIXED
 
 Dr said Cash Card Cheque and Slip enough for payment methods
 
 
 Income Details- There were two Payment Methods text amounts should be renamed ----> FIXED
 
 Card-> Last 4 digits-> can enter more than number and less than number need validation ----> Will do with a new issue


Pubudu-Piyankara (approved):
reviewed

PR #8267, opened 2024-10-25
Issue#8245

Pubudu-Piyankara:
When an OPD bill is settle using the Patient Deposit Payment Method, it is not updated in the drawer.

Total values are not considering the patient deposit amount now.


DamithDeshan (approved):
No Issues

PR #8795, opened 2024-11-14
Issue#8794

Pubudu-Piyankara:
supplier payment does not need comment,therefore I have removed the navigation to reprint_bill_dealor and add approve
method to approve button with validation.
/dealerPayment/search_dealor_payment.xhtml


PasinduW99 (approved):
Done

PR #8973, opened 2024-11-22
Issue#8448

Pubudu-Piyankara:
Need comment for Shift End IS DONE
cashier/shift_end_for_handover.xhtml

There is a application option to this as "Allow Comment for Shift End"
There is a application option to this as "Allow denomination for Shift Ending process"


PasinduW99 (approved):
Done

Issue #8991, opened 2024-11-22
Error in Institution and Department in CC Billing


buddhika75:
Error in Institution and Department in CC Billing

Currently, when a CC bill is done, the CC and its department are selected for the bills department and bills institution. But is should be the Users Logged Institution and Department that is recorded as the institution and Department.

buddhika75:
Up to now, reports were checked for logged users institution to check the reports of CC Bills, But this fails if the user changes the Institution or Department. This will affect the Bills after this update is deployed as production. For past data, to correctly work for the institution and Department, we need to change the CC Bills department from Collection Centres.

buddhika75:
Dear Team,

Regarding the issue titled "Error in Institution and Department in CC Billing" (#8991) on GitHub:

The issue has been identified and a fix has been implemented. The key updates are as follows:

Current Issue: Previously, CC Bills recorded the Collection Centre (CC) and its department as the bill's institution and department. However, it should reflect the logged user's institution and department.

Resolution: The logic has been updated to use the logged user's institution and department as the CC Bill's institution and department instead of the CC. Currently, the Collection Centre is already recorded as CollectingCentre and fromInstitution, and these fields remain unchanged. As a result, reports searched by CC are unaffected.

Impact on Reports:

Reports filtering by institution, department, or site currently work as they rely on the logged user's institution and department.
If the user is transferred to another institution or site, reports with these filters may not work as intended for past data.
For all new CC Bills created after the update is deployed to production, filtering by institution and department will work without any issue.
Clarification on Practical Impact:

Searching with the Collection Centre or route remains unaffected.
Practical issues with past data only arise if a single Collection Centre sends samples to multiple institutions or sites.
For institutions like Ruhunu Hospital, this has no significant implications if reports are searched without specifying the billed institution, site, or department.
Next Steps: Please ensure this is communicated to relevant stakeholders and prepare for any adjustments needed post-deployment.

Let me know if further clarification is required.

Best regards,
Buddhika

PR #9001, opened 2024-11-23
Issue#8991

buddhika75:
Dear Team,

Regarding the issue titled "Error in Institution and Department in CC Billing" (#8991) on GitHub:

The issue has been identified and a fix has been implemented. The key updates are as follows:

Current Issue: Previously, CC Bills recorded the Collection Centre (CC) and its department as the bill's institution and department. However, it should reflect the logged user's institution and department.

Resolution: The logic has been updated to use the logged user's institution and department as the CC Bill's institution and department instead of the CC. Currently, the Collection Centre is already recorded as CollectingCentre and fromInstitution, and these fields remain unchanged. As a result, reports searched by CC are unaffected.

Impact on Reports:

Reports filtering by institution, department, or site currently work as they rely on the logged user's institution and department.
If the user is transferred to another institution or site, reports with these filters may not work as intended for past data.
For all new CC Bills created after the update is deployed to production, filtering by institution and department will work without any issue.
Clarification on Practical Impact:

Searching with the Collection Centre or route remains unaffected.
Practical issues with past data only arise if a single Collection Centre sends samples to multiple institutions or sites.
This has no significant implications if reports are searched without specifying the billed institution, site, or department.
Next Steps: Please ensure this is communicated to relevant stakeholders and prepare for any adjustments needed post-deployment.

Let me know if further clarification is required.

Best regards,
Buddhika


DamithDeshan (approved):
No Issues

PR #9163, opened 2024-11-28
Issue#8846

Pubudu-Piyankara:
When Select the Credit Payment Method, Error page was loaded in Direct Purchase >>> SOLVED

Pharmacy/Procurements/Direct Purchase


PasinduW99 (approved):
Done

PR #9267, opened 2024-12-02
Issue#9252

ChinthakaPrasad:
Fix: when create booking cant save without selecting department. Change few updates in channel booking by date ui.


SandithP2001 (approved):
Reviewed

PR #9269, opened 2024-12-02
Issue#9263

Pubudu-Piyankara:
Supplier payment column filters need to update

/dealerPayment/dealor_due.xhtml

Supplier payment management > payment to settle

column header filter does not work

DONE


SandithP2001 (approved):
Reviewed

PR #9433, opened 2024-12-06
Issue#9432

SandithP2001:
Cancellation Path: Pharmacy/Bill Search -> Find Bill & Cancel

Doc No.

Ref Doc No. (Cancellation Doc No not updated)

Ref Doc Date (Cancellation Date not updated)

Document Type wrong (Currently Displayed as 'Pharmacy Retail Sale Cancelled') - To fix this need to change the bill type atomic of the direct purchase bil cancelation method. But I think the same method may be used in many cancellation methods. So need to check discuss and fix. I asked from dr and he told me to change that. But need to check.

Others are corrected.


DamithDeshan (approved):
No Issues

PR #9437, opened 2024-12-06
Issue#9305

Pubudu-Piyankara:
ward/ward_pharmacy_bht_issue.xhtml

 Alreaady >>> Need validation if Qty was 0 or more than qty (when added more than qty, it is set to zero)
 Issue to BHT- cannot issue 0 or more than qty


SandithP2001 (approved):
Reviewed

PR #9446, opened 2024-12-07
Issue#9305

Pubudu-Piyankara:
ward/ward_pharmacy_bht_issue.xhtml

 Need validation if Qty was 0 or more than qty
 Issue to BHT- cannot issue 0 or more than qty
 
 if Qty is set to  more than available stock it is set to zero by default.


DamithDeshan (approved):
No Issues

PR #9542, opened 2024-12-10
Issue#9386 opd package bills are possible under patient points even if there is no patient deposits

buddhika75:


DamithDeshan (approved):
No Issues

PR #9597, opened 2024-12-13
Issue#9595 set Institution in pharmacy transfer issue bill

Pubudu-Piyankara:
Need to set Institution in pharmacy transfer issue bill


DamithDeshan (approved):
No Issues

PR #9620, opened 2024-12-13
Issue#9404 closing stock

SandithP2001:
Item wise and Batch wise reports are done. There is a closing stock value issue it will fix with another issue.


DamithDeshan (approved):
No Issues

PR #9669, opened 2024-12-17
9457 improvements in phamacy sale add to stock

ChinthakaPrasad:
pharmacy sale add to stock improvement and sale for cashier no need to enter amount when billing.


SandithP2001 (approved):
Reviewed

PR #9724, opened 2024-12-19
9456 phamacy disbursement and procument improvements

ChinthakaPrasad:
Generate supplier code functionality is done.
(admin -> pharmaceutical management -> institution -> supplier)
Add to stock date range filtering and multiple selection is done.
(pharmacy->retail -> Add to stock)
Pharmacy bill return from item and payment return section drawer update is done.
(Pharmacy->retail->return item and payment)


DamithDeshan (approved):
No Issues

Issue #6674, opened 2024-07-28
External Doctor for Collection Center Billing


DeshaniPubudu:
External Doctor for Collection Center Billing 

(With Only add doctors if they have submitted all the details of doctor with SLMC. For CC Bills they need to feed Referring Doctor Name is not Available in the system they need to feed that doctor name in the CC Invoice. That Doctor name no need to save in our consultant list.)

DamithDeshan:
### **Worning !**
This External Doctor use **Only Bill**.
So this Dorctor name can't use to any Reports

PR #9793, opened 2024-12-20
#6674 add external doctor for cc billing

DamithDeshan:


DamithDeshan:
### **Worning !**
This External Doctor use **Only Bill**.
So this Dorctor name can't use to any Reports

SandithP2001 (approved):
Reviewed

Issue #9855, opened 2024-12-24
When searching for a members' phone number directly through the pharmacy retail sale, membership is not filled in by default.


DeshaniPubudu:
Without going through the membership,
When searching for a phone number related to a Member through the Pharmacy Retail Sales, the member's membership should be filled in by default.

Pharmacy/Retail Sale

buddhika75:
Done

PR #9896, opened 2024-12-28
When searching for a members' phone number directly through the pharmacy retail sale, membership is not filled in by default. #9855

buddhika75:


DamithDeshan (approved):
No Issues

PR #10005, opened 2025-01-06
10004 improvements in expiry item report

SandithP2001:


SandithP2001:
Cannot add columns - parent category code, parent category name, secondary UOM ---> these values not record in our system
Batch wise total, Batch wise Qty, Item wise total, Item wise Qty ----> Can get values for this columns but it will repeat same value in many rows (Dr said need to discuss with Dr)

ChinthakaPrasad (approved):

PR #10056, opened 2025-01-09
9994 search filter issue in inpatient direct issue by bill

ChinthakaPrasad:
supplier upload validation is fixed for empty cells.
Direct issue for BHT bill search by department is done.

### DIFF
--- a/src/main/java/com/divudi/bean/emr/DataUploadController.java
+++ b/src/main/java/com/divudi/bean/emr/DataUploadController.java
@@ -4623,6 +4623,19 @@ public class DataUploadController implements Serializable {
 
         return agencyList;
     }
+    
+    public boolean isRowEmpty(Row row){
+        if(row == null){
+            return true;
+        }
+        
+        for(Cell cell : row){
+            if(cell != null && cell.getCellType()!= CellType.BLANK){
+                return false;
+            }
+        }
+        return true;
+    }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/divudi/bean/emr/DataUploadController.java
```diff
@@ -4623,6 +4623,19 @@ private List<Institution> readAgenciesFromExcel(InputStream inputStream) throws
 
         return agencyList;
     }
+    
+    public boolean isRowEmpty(Row row){
+        if(row == null){
```
GSMgeeth:
``` Java
if (row != null) {
    for (Cell cell : row) {
        if (cell != null && cell.getCellType() != CellType.BLANK) {
            return false;
        }
     }
}

return true;
```

src/main/java/com/divudi/bean/emr/DataUploadController.java
```diff
@@ -4638,6 +4651,10 @@ private List<Institution> readSuppliersFromExcel(InputStream inputStream) throws
         }
         while (rowIterator.hasNext()) {
             Row row = rowIterator.next();
+            
+            if(row == null || isRowEmpty(row)){
```
GSMgeeth: The `isRowEmpty()` method already checks row nullability. You can ignore it here and only do `if (isRowEmpty(row))`.

GSMgeeth (changes requested):

PR #10110, opened 2025-01-15
#9409 stock transfer report

Pubudu-Piyankara:
/reports/inventoryReports/stock_transfer_report.xhtml

Stock Transfers Report >>> DONE

### DIFF
--- a/src/main/java/com/divudi/bean/pharmacy/PharmacyController.java
+++ b/src/main/java/com/divudi/bean/pharmacy/PharmacyController.java
@@ -1637,7 +1637,7 @@ public class PharmacyController implements Serializable {
         }
     }
 
-    private void additionalCommonFilltersForBillEntity(StringBuilder sql, Map<String, Object> parameters ) {
+    private void additionalCommonFilltersForBillEntity(StringBuilder sql, Map<String, Object> parameters) {
         addFilter(sql, parameters, "b.fromInstitution", "fIns", fromInstitution);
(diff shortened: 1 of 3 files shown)


src/main/java/com/divudi/bean/pharmacy/PharmacyController.java
```diff
@@ -1585,18 +1585,156 @@ public void createStockTransferReport() {
...
+        // Adding the ordering clause
+        sql.append(" ORDER BY bi.id DESC");
+
+        // Executing the query
+        try {
+            billItems = getBillItemFacade().findByJpql(sql.toString(), parameters, TemporalType.TIMESTAMP);
+        } catch (Exception e) {
+            FacesContext.getCurrentInstance().addMessage(null,
+                    new FacesMessage(FacesMessage.SEVERITY_ERROR, "Error", "Failed to generate report. Please try again."));
+            e.printStackTrace();
```
GSMgeeth, line 1683: This method takes a lot of resources, pollute logs, and can add up to making large size log files. Printing the `e` with necessary error level might be enough in most cases. Sometimes even only printing the message as `e.getMessage()` with error level is enough for the deployed version. Current method can be useful when debugging locally but should not commit.

Also, better to use a logging framework / library.

**Example:**
```Java
try {
    // Code that might throw an exception
} catch (Exception e) {
    logger.error("An error occurred while processing", e);
}
```

GSMgeeth (approved):

PR #10112, opened 2025-01-15
10102 shift start contains handed over values in the previous shift

buddhika75:
Now when handing over completed to another user, they do not appear in the summaries for the next shift.

Please check @DeshaniPubudu


src/main/java/com/divudi/bean/cashTransaction/FinancialTransactionController.java
```diff
@@ -2404,20 +2404,24 @@ public String navigateToCreateShiftEndSummaryBillForHandover() {
             shiftPayments.stream()
                     .forEach(p -> p.setTransientPaymentHandover(PaymentHandover.USER_COLLECTED));
         }
+        System.out.println("shiftPayments = " + shiftPayments);
```
GSMgeeth, line 2407: Better to either **remove before committing** or **use a logging library / framework to log** information or errors. 

The `System.out.println("")` method can cause performance issues and also not designed for thread-safety among other weaknesses when compared to a logging framework.

GSMgeeth (approved):

PR #11234, opened 2025-03-20
Core module separation

GSMgeeth:
- Refactor CommonFunctions util class
- Moves util classes into /java/ package (to be renamed later)
- Moves entity class responsibilities into relevant entity classes
- Other refactoring


buddhika75 (approved):
No Issue

PR #11257, opened 2025-03-21
Core module separation

GSMgeeth:


buddhika75 (approved):
All ok. Only double about making Double to double in report template row class. May be in jsf pages, we want to display a blank space instead of a 0.00 in some reports. WIll create an issue again if such instance is found in future.

PR #11320, opened 2025-03-25
Core module separation

GSMgeeth:


src/main/java/com/divudi/ws/channel/ChannelApi.java
```diff
@@ -927,7 +906,7 @@ public Response createBooking(@Context HttpServletRequest requestContext, Map<St
             msgForStartSession = "Session is starting now. Please complete booking and visit quickly";
         }
 
-        if (patientDetails == null || patientDetails.isEmpty()) {
+        if (patientDetails.isEmpty()) {
```
buddhika75, line 909: Null check was removed. This will cause a NullPointerException if patientDetails is null.
Recommend restoring: if (patientDetails == null || patientDetails.isEmpty())

buddhika75 (changes requested):
Null check was removed. This will cause a NullPointerException if patientDetails is null.
Recommend restoring: if (patientDetails == null || patientDetails.isEmpty())

src/main/java/com/divudi/ws/channel/ChannelApi.java
GSMgeeth: In this line, the `patientDetails` cannot be null due to a check above (L: 856)

PR #11358, opened 2025-03-26
Report monitoring

ImeshRanawellaSG:


src/main/java/com/divudi/bean/common/ReportTimerController.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        if (reportLog.getId() == null) {
+            reportLog.setCreatedAt(new Date());
+            reportLog.setGeneratedBy(sessionController.getLoggedUser());
+
+            try {
+                getFacade().create(reportLog);
+            } catch (Exception e) {
+                getFacade().edit(reportLog);
```
GSMgeeth: In this line, the reportLog.getId() will be `null`. So, is edit() the correct way to handle this exception?

src/main/java/com/divudi/bean/common/ReportTimerController.java
```diff
@@ -0,0 +1,69 @@
...
+        save(reportLog);
+    }
+
+    public void save(ReportLog reportLog) {
+        if (reportLog == null) {
+            return;
+        }
+
+        if (reportLog.getId() == null) {
+            reportLog.setCreatedAt(new Date());
```
GSMgeeth: The ReportLog object itself update the createdAt value when creating the object. Is this needed?

src/main/java/com/divudi/entity/report/ReportLog.java
```diff
@@ -0,0 +1,173 @@
...
+
+    public ReportLog setEndTime(Date endTime) {
+        this.endTime = endTime;
+
+        return this;
+    }
+
+    public ReportLog setReport(IReportType reportType) {
+        this.reportType = reportType.getReportType();
+        this.reportName = reportType.getValue();
```
GSMgeeth: Should this be `reportType.getDisplayName()` ?

src/main/java/com/divudi/data/reports/IReportType.java
```diff
@@ -0,0 +1,11 @@
+package com.divudi.data.reports;
+
+public interface IReportType {
+    String getDisplayName();
+
+    String getReportPath();
```
GSMgeeth: Let's add this later only if necessary

src/main/java/com/divudi/data/reports/CollectionCenterReport.java
```diff
@@ -0,0 +1,33 @@
+package com.divudi.data.reports;
+
+public enum CollectionCenterReport implements IReportType {
+    ROUTE_ANALYSIS_REPORT("Route Analysis Report", "route_analysis_report.xhtml");
+
+    private final String displayName;
+    private final String reportPath;
```
GSMgeeth: Let's keep report path only when necessary.

src/main/java/com/divudi/bean/common/ReportTimerController.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Date;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+@Stateless
+public class ReportTimerController implements Serializable {
+    private static final Logger LOGGER = Logger.getLogger(ReportTimerController.class.getName());
+
+    @Inject
+    SessionController sessionController;
```
GSMgeeth: Let's get rid of this dependency and accept logged user as a parameter. All report generating classes will already have this as a dependency.

src/main/java/com/divudi/bean/common/ReportTimerController.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    public void save(ReportLog reportLog) {
+        if (reportLog == null) {
+            return;
+        }
+
+        if (reportLog.getId() == null) {
+            reportLog.setCreatedAt(new Date());
+            reportLog.setGeneratedBy(sessionController.getLoggedUser());
```
GSMgeeth: When logged user is passed as a parameter, we can set logged user in above method. With that, consider converting ReportLog to a simple Value-like final class with final private attributes (immutable object). I think value objects are not supported in this Java version to be used directly. Double check on how this change will affect when persisting and fetching from database.

src/main/java/com/divudi/bean/common/ReportTimerController.java
```diff
@@ -0,0 +1,69 @@
...
+
+        try {
+            reportGenerationLogic.run();
+        } catch (Exception e) {
+            LOGGER.log(Level.SEVERE, "Error occurred while generating the report", e);
+        }
+
+        final Date endTime = new Date();
+
+        final long durationInMilliseconds = endTime.getTime() - startTime.getTime();
```
GSMgeeth: This maybe suitable inside setStartTime() and setEndTime() methods. When one of them is updated, automatically duration will be calculated and assigned. A check will be needed to make sure both fields are not null before updating.

src/main/java/com/divudi/bean/common/ReportTimerController.java
```diff
@@ -0,0 +1,69 @@
...
+import javax.ejb.EJB;
+import javax.ejb.Stateless;
+import javax.inject.Inject;
+import java.io.Serializable;
+import java.util.Date;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+@Stateless
+public class ReportTimerController implements Serializable {
```
GSMgeeth, line 17: Once the class is free from `SessionController`, consider moving this to /java/ package with a proper name. Package /java/ is to be renamed as /util/ soon. Facade dependency is okay.

GSMgeeth (changes requested):

src/main/java/com/divudi/bean/common/ReportTimerController.java
ImeshRanawellaSG: JPA does not support final entity classes because Hibernate and other JPA providers need to create proxy subclasses for lazy loading and other features.

GSMgeeth (approved):

PR #11757, opened 2025-04-16
11756 need pharmacy retail sale and inpatient sale price with purchase values hotfix

buddhika75:
get new report urgently


buddhika75:
Do NOT merge. Its a mistake. Updating from DIV

PR #13109, opened 2025-06-16
12709 direct issue bill value should not be calculated by sale ratebill value should be calculated using purchase rate

buddhika75:
done

## Summary by CodeRabbit

- **New Features**
  - Enhanced bill financial details to support cost rate-based calculations for pharmacy bills, controlled by a configurable option.

- **Style**
  - Updated the pharmacy transfer page layout for improved structure and visual emphasis, including new panel styling and repositioned forms.


buddhika75:
double click and check
also open 2 windoes - settle in one first, then try to settle again in the second window.

PR #15543, opened 2025-09-23
15321 stock ledger   department filter does not filter out stocks for department

buddhika75:
## Summary by CodeRabbit

* New Features
  * Reports show closing stock/value per Department, Institution, or Total with badges and include cost-rate metrics.
  * Stock history now records per-item, institutional and total stock and their value at purchase and cost rates.
* Bug Fixes
  * Existing issued bills are properly updated when saved.
  * Creation metadata is preserved when already set.
* Refactor
  * Print behavior now targets the table without post-print restoration.
* Data/Reports
  * Bill export/report columns reordered and expanded.


buddhika75:
@coderabbitai - can you please check now

PR #15719, opened 2025-10-03
15716-need-ui-improvement-for-pharmacy-payment-by-credit-company

ishira-web:
Put a Space between buttons 
 Remove Date and Time and Set it into One Line

### DIFF
--- a/src/main/webapp/credit/credit_compnay_bill_pharmacy_all.xhtml
+++ b/src/main/webapp/credit/credit_compnay_bill_pharmacy_all.xhtml
@@ -214,7 +214,8 @@
                         ajax="false" 
                         action="#{cashRecieveBillController.prepareNewBill()}"  
                         icon="fa fa-plus"
-                        class="ui-button-warning">
+                        class="ui-button-warning"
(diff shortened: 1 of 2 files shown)


src/main/webapp/credit/credit_compnay_bill_pharmacy_all.xhtml
```diff
@@ -214,7 +214,8 @@
                         ajax="false" 
                         action="#{cashRecieveBillController.prepareNewBill()}"  
                         icon="fa fa-plus"
-                        class="ui-button-warning">
+                        class="ui-button-warning"
+                        style="margin-right: 20px;">
```
buddhika75, line 218: Do not use inline styles. Please use Bootstrap classes like me-1

src/main/webapp/resources/bill/insRecieve.xhtml
buddhika75, line 1: Please use a new bill for Cancel Bill. Keep the original one for printing at the time of billing. For cancellation, create a new print with these details.

Issue #15869, opened 2025-10-14
Cost Of Good Sold-GRN Cash


Irani96:
purchase value of the Cash GRN is incorrect in the cost of Good sold report.
https://drive.google.com/open?id=1d_moq9WOFb63biXN2nZ6nhD-kMMPoUVl
https://drive.google.com/open?id=1w2eDuobh-b9_cnewqIMMMWFiQ_Pop7i-
File Path: reports/inventoryReports/grn_cash.xhtml

PR #15955, opened 2025-10-17
Cost Of Good Sold-GRN Cash #15869

Pubudu-Piyankara:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the Purchase Value calculation in the GRN Cash inventory report to accurately reflect the product of purchase rate and quantity, ensuring correct values are displayed in the report.

### DIFF
--- a/src/main/webapp/reports/inventoryReports/grn_cash.xhtml
+++ b/src/main/webapp/reports/inventoryReports/grn_cash.xhtml
@@ -300,7 +300,7 @@
                                 </h:outputText>
                             </p:column>
                             <p:column headerText="Purchase Value" styleClass="text-end" headerStyleClass="text-end" width="11em">
-                                <h:outputText value="#{bi.billItemFinanceDetails.valueAtPurchaseRate}">
+                                <h:outputText value="#{bi.billItemFinanceDetails.purchaseRate * bi.billItemFinanceDetails.quantity}">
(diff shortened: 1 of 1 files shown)


src/main/webapp/reports/inventoryReports/grn_cash.xhtml
```diff
@@ -300,7 +300,7 @@
                                 </h:outputText>
                             </p:column>
                             <p:column headerText="Purchase Value" styleClass="text-end" headerStyleClass="text-end" width="11em">
-                                <h:outputText value="#{bi.billItemFinanceDetails.valueAtPurchaseRate}">
```
buddhika75: This should work for new bills. Old bills need to have their data updated, as the calculation was done recently. If not working, we have to fix it at the point of GRN Creation.

src/main/webapp/reports/inventoryReports/grn_cash.xhtml
```diff
@@ -300,7 +300,7 @@
...
+                                <h:outputText value="#{bi.billItemFinanceDetails.purchaseRate * bi.billItemFinanceDetails.quantity}">
```
buddhika75: Purchase Value in GRN should include the free quantities depending on the preference. It is should therefore NOT be calculated here.

PR #18751, opened 2026-02-22
18414 pharmacy purchase rate adjustment not updated in f15

buddhika75:
Bug 1 — Duplicate persist (double create/edit of BillItem):
  The bill item was persisted at lines 1038-1046, then again at lines 1087-1091 after the BFD updates. Fixed to a single
   create call.

  Bug 2 — Wrong quantity variable for before/after values:
  beforeVal and afterVal used getStock().getStock() (the stock field on the controller, i.e., the last batch assigned in
   the loop) instead of dto.getStockQty(). For multi-batch adjustments this picked the wrong quantity. Fixed to use
  dto.getStockQty().

  Bug 3 — BFD fields overwritten instead of accumulated (multi-batch):
  bfd.setNetTotal(changeVal), bfd.setGrossTotal(...), and bfd.setTotalQuantity(...) used assignment instead of prev +
  current. For multi-batch adjustments only the last batch's values survived. Fixed to accumulate with prev.add(...).

  Bug 4 — Zeroing out already-accumulated values:
  bfd.setTotalCostValue(ZERO), setTotalRetailSaleValue(ZERO), setTotalWholesaleValue(ZERO) unconditionally reset to zero
   on every iteration, wiping out any previously accumulated purchase value too (since BFD is shared). Changed to only
  initialize if null, which is the correct intent (purchase rate adjustments don't affect cost/retail/wholesale).

  Bug 5 — In-memory bill items list not updated:
  getDeptAdjustmentPreBill().getBillItems().add(tbi) was commented out, causing incorrect serial number sequencing for
  subsequent items. Uncommented.

## Summary by CodeRabbit

* **Bug Fixes**
  * More consistent and accurate pharmacy bill adjustment calculations across all adjustment types.
  * Improved initialization and null-safe handling of financial details to prevent missing or incorrect totals.
  * Corrected batch-level inventory valuation to ensure accurate before/after values.
  * Removed legacy handling paths to stabilize item creation and total propagation in adjustments.

### DIFF
--- a/src/main/java/com/divudi/bean/pharmacy/PharmacyAdjustmentController.java
+++ b/src/main/java/com/divudi/bean/pharmacy/PharmacyAdjustmentController.java
@@ -1184,6 +1184,7 @@ public class PharmacyAdjustmentController implements Serializable {
             Logger.getLogger(PharmacyAdjustmentController.class.getName()).log(Level.SEVERE, "Failed to save cost rate adjustment bill items", e);
             throw new RuntimeException("Failed to save cost rate adjustment bill items", e);
         }
+        getDeptAdjustmentPreBill().getBillItems().add(tbi);
 
         Double currentTotalObj = getDeptAdjustmentPreBill().getTotal();
         Double currentNetTotalObj = getDeptAdjustmentPreBill().getNetTotal();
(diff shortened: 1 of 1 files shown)


buddhika75:
@coderabbitai review

The CodeRabbit review identified a valid bug in `saveCrAdjustmentBillItems`: BFD fields were being overwritten instead of accumulated across multiple batch iterations, there was a duplicate persist block, and `getBillItems().add(tbi)` was missing. This has been fixed in the latest commit (86cd837):

- Replace `bfd.setTotalCostValue`, `bfd.setNetTotal`, `bfd.setGrossTotal`, `bfd.setTotalQuantity` overwrites with accumulation using `prev.add()` pattern
- Change unconditional zero-sets for `totalPurchaseValue`, `totalRetailSaleValue`, `totalWholesaleValue` to null-check only
- Remove duplicate `getBillItemFacade().create/edit()` block at end of method
- Add `getDeptAdjustmentPreBill().getBillItems().add(tbi)` for correct serial numbering

Please re-review.

PR #19042, opened 2026-03-06
Users can now cancel a created GRN Return directly from the fi…

ishira-web:
- A Cancel button shows up in the Actions column for any GRN Return that hasn't been cancelled or finalized yet. Click it, confirm the dialog (which shows the return number), and it's done .you stay on the
  same page and the table refreshes in place.
  - Every cancellation gets written to the audit log with a proper before/after snapshot, so you can see exactly what changed, who did it, and when.

## Summary by CodeRabbit

* **New Features**
  * Cancel GRN Returns directly from the list via a new Cancel button with confirmation; option hidden for cancelled or completed returns.

* **Improvements**
  * Action column layout adjusted for better spacing and responsive wrapping.
  * Audit logging improved to capture reliable before/after states for cancellations.

* **Bug Fixes**
  * Audit handling made more robust when stored states are not JSON objects.

* **UI Cleanup**
  * Removed redundant cancelled-status display blocks from list view.

### DIFF
--- a/src/main/java/com/divudi/bean/common/AuditEventApplicationController.java
+++ b/src/main/java/com/divudi/bean/common/AuditEventApplicationController.java
@@ -13,6 +13,8 @@ import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import javax.ejb.EJB;
+import java.util.logging.Level;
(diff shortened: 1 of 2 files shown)


ishira-web:
**Audit Log**

image

buddhika75 (approved):

PR #19866, opened 2026-04-09
Replace jasypt URL tokens with HMAC-SHA256 for bill notification links

### LINKED ISSUE #19863: Security hardening: fix critical vulnerabilities and reduce patient data exposure risk
## Background

A security review was conducted against the REST API layer, authentication subsystem, and web configuration to identify risks of unauthorized patient data extraction. This issue tracks all findings and the remediation plan.

---

### 1. Authentication backdoor — `SecurityController.matchPassword(String, String, boolean)` always returns `true`

**File:** `src/main/java/com/divudi/bean/common/SecurityController.java:180`

```java
public static boolean matchPassword(String planePassword, String encryptedPassword, boolean fake) {
    return true;  // ← unconditionally bypasses authentication
}
```

This overload exists alongside the real `matchPassword(String, String)`. Any caller that accidentally passes a third argument (boolean) will bypass password verification entirely. Even if unused today, it is a latent critical risk and must be removed.

**Action:** Delete this overload.

---

### 2. Hardcoded encryption key `"health"`

**File:** `src/main/java/com/divudi/bean/common/SecurityController.java:143,197`

```java
en.setPassword("health");
```

Both `encrypt()` and `decrypt()` use a hardcoded, trivially guessable key. Any data protected by this encryption is effectively plaintext to anyone who reads the source.

**Action:** Move the key to a Payara JNDI config entry or environment variable. Rotate any data encrypted under the old key.

---

### 3. CORS misconfiguration — wildcard origin with credentials

**File:** `src/main/java/com/divudi/ws/channel/CorsResponseFilter.java:21-24`

```java
responseContext.getHeaders().add("Access-Control-Allow-Origin", "*");
responseContext.getHeaders().add("Access-Control-Allow-Credentials", "true");
```

`*` + `credentials: true` is invalid per the CORS spec (browsers reject it) but signals intent to allow all origins. The correct fix is to allowlist only trusted origins.

**Action:** Replace `*` with an explicit allowlist of known frontend origins, read from configuration.

---

### 4. Weak password hashing — `BasicPasswordEncryptor` (MD5-based)

**File:** `src/main/java/com/divudi/bean/common/SecurityController.java:161`

Jasypt's `BasicPasswordEncryptor` uses MD5 with a salt. MD5 is fast and GPU-crackable. For a healthcare system storing patient data, bcrypt or Argon2 should be used.

**Action:** Migrate to `StrongPasswordEncryptor` (Jasypt) which uses SHA-256 with 1000 iterations, or switch to bcrypt. Plan a one-time password re-hash on next login for existing accounts.

---

### 5. No rate limiting on REST API endpoints

No throttling or brute-force protection exists on any `/api/*` endpoint, including login and user management APIs.

**Action:** Add a simple request counter filter (per IP or per API key) that returns HTTP 429 after a configurable threshold. Payara's built-in request filtering or a `ContainerRequestFilter` can handle this.

---

### 6. Session timeout too long for healthcare data

**File:** `src/main/webapp/WEB-INF/web.xml:61`

```xml
<session-timeout>180</session-timeout> <!-- minutes -->
```

3-hour idle sessions are excessive for a system with access to patient billing, clinical, and pharmacy data.

**Action:** Reduce to 30–60 minutes, with a configurable override via `ConfigOption`.

---

### 7. Error pages may leak stack traces

**File:** `src/main/webapp/WEB-INF/web.xml:87-89`

`java.lang.Throwable` is caught generically. Ensure the error pages (`generalError.xhtml`, `500.xhtml`) never render exception messages, class names, or stack traces to the browser.

**Action:** Audit all error XHTML pages to confirm no `#{facesContext.maximumSeverity}` or raw exception output is rendered.

---

## Informational

- **API key header name `Finance`** — non-standard but functional; document it clearly so it is not mistaken for a finance-domain header.
- **Source code visibility** — making the repo private adds minimal security value. The real risks are the issues above; address those first.

---

## Remediation Priority

| # | Issue | Severity | Effort |
|---|-------|----------|--------|
| 1 | Delete `matchPassword(,,boolean)` backdoor | Critical | 5 min |
| 2 | Externalize hardcoded encryption key | Critical | Medium |
| 3 | Fix CORS to allowlist origins | High | Small |
| 4 | Upgrade password hashing to bcrypt | High | Medium |
| 5 | Add REST API rate limiting | High | Medium |
| 6 | Reduce session timeout | Medium | Small |
| 7 | Audit error pages for info leakage | Medium | Small |

---

## References
- OWASP Top 10: A02 Cryptographic Failures, A05 Security Misconfiguration, A07 Identification and Authentication Failures
- HIPAA Technical Safeguards: Access Control, Audit Controls, Transmission Security

buddhika75:
## Summary

- Bill notification URLs (OPD cancellation emails/SMS) were signed with a hardcoded Jasypt symmetric key (`"health"`), making tokens forgeable by anyone who reads the source code
- Replaces encrypt/decrypt pair with **HMAC-SHA256** tokens signed by the per-installation key already stored in `ApplicationPreference.encrptionKey`
- New token format: `hmac.{billId}.{expiryEpochMs}.{base64url_signature}` — self-expiring, tamper-proof, no separate `&user=` expiry param needed
- No database migration required; the signing key already exists in every installation

## Files changed

| File | Change |
|---|---|
| `SecurityController.java` | Added `createBillToken()`, `decodeBillToken()`, `computeHmacSha256()` |
| `NotificationController.java` | Both OPD cancellation notification methods now use `createBillToken()` |
| `BillSearch.java` | `preparePatientReportByIdForRequests()` — HMAC path + legacy fallback |
| `PatientReportController.java` | `preparePatientReportByIdForRequests()` — same pattern |

## Backward compatibility

Old jasypt-encrypted links already sent are still valid via a legacy fallback block in both receivers. Both blocks carry:

```java
// TODO: Remove this legacy block after 2026-07-09.
// Backward compatibility for links sent before HMAC migration (issue #19863).
// Old links have a 1-month TTL so none will be valid after that date.
```

## Test plan

- [ ] Send an OPD bill cancellation notification and verify the link in the email/SMS opens the correct bill
- [ ] Verify an expired HMAC token (manually set past expiry) shows "No Such Bill Found"
- [ ] Verify a tampered token (modify a character in the signature) shows "No Such Bill Found"
- [ ] Confirm old-format links (jasypt-encrypted, `?id=...&user=...`) still work during the transition window

Closes #19863

## Summary by CodeRabbit

* **Security**
  * Added HMAC-signed tokens for bill/report access links with expiry and signature validation.
  * Improved signing key management and safer token decoding.

* **Bug Fixes / UX**
  * Links (email/QR/cancel) now carry a single opaque token parameter, reducing exposed query parameters.

* **Refactor**
  * Dual support for new HMAC tokens and legacy links to preserve compatibility.

### DIFF
--- a/src/main/java/com/divudi/bean/common/BillSearch.java
+++ b/src/main/java/com/divudi/bean/common/BillSearch.java
@@ -626,34 +626,48 @@ public class BillSearch implements Serializable, ControllerWithMultiplePayments 
         if (encryptedPatientReportId == null) {
             return;
         }
-        if (encryptedExpiary != null) {
-            Date expiaryDate;
-            try {
-                String ed = encryptedExpiary;
-                ed = securityController.decrypt(ed);
-                if (ed == null) {
+        if (encryptedPatientReportId.startsWith("hmac.")) {
+            // New HMAC-SHA256 path — token embeds both bill ID and expiry
+            String hmacKey = sessionController.getApplicationPreference().getEncrptionKey();
+            if (hmacKey == null || hmacKey.trim().isEmpty()) {
+                return;
+            }
+            long[] decoded = securityController.decodeBillToken(encryptedPatientReportId, hmacKey);
+            if (decoded == null) {
+                return;
+            }
+            if (new Date().getTime() > decoded[1]) {
+                return; // link expired
+            }
+            bill = getBillFacade().find(decoded[0]);
+        } else {
+            // TODO: Remove this legacy block after 2026-07-09.
+            // Backward compatibility for links sent before HMAC migration (issue #19863).
+            // Old links have a 1-month TTL so none will be valid after that date.
+            if (encryptedExpiary != null) {
+                Date expiaryDate;
+                try {
+                    String ed = securityController.decrypt(encryptedExpiary);
+                    if (ed == null) {
+                        return;
+                    }
+                    expiaryDate = new SimpleDateFormat("ddMMMMyyyyhhmmss").parse(ed);
+                } catch (ParseException ex) {
                     return;
                 }
-                expiaryDate = new SimpleDateFormat("ddMMMMyyyyhhmmss").parse(ed);
-            } catch (ParseException ex) {
+                if (expiaryDate.before(new Date())) {
+                    return;
+                }
+            }
+            String idStr = getSecurityController().decrypt(encryptedPatientReportId);
+            Long id = 0L;
+            try {
+                id = Long.parseLong(idStr);
+            } catch (Exception e) {
                 return;
             }
-            if (expiaryDate.before(new Date())) {
-                return;
-            }
+            bill = getBillFacade().find(id);
         }
-        String idStr = getSecurityController().decrypt(encryptedPatientReportId);
-        Long id = 0l;
-        try {
-            id = Long.parseLong(idStr);
-        } catch (Exception e) {
-            return;
-        }
-        Bill pr = getBillFacade().find(id);
-        if (pr == null) {
-            return;
-        }
-        bill = pr;
     }
 
     public void fillBillTypeSummery() {
--- a/src/main/java/com/divudi/bean/common/NotificationController.java
+++ b/src/main/java/com/divudi/bean/common/NotificationController.java
@@ -277,20 +277,20 @@ public class NotificationController implements Serializable {
         Calendar c = Calendar.getInstance();
         c.add(Calendar.MONTH, 1);
 
-        String temId = bill.getId() + "";
-        temId = securityController.encrypt(temId);
-        try {
-            temId = URLEncoder.encode(temId, "UTF-8");
-        } catch (UnsupportedEncodingException ignored) {
+        String hmacKey = sessionController.getApplicationPreference().getEncrptionKey();
+        if (hmacKey == null || hmacKey.trim().isEmpty()) {
+            hmacKey = securityController.generateRandomKey(32);
+            sessionController.getApplicationPreference().setEncrptionKey(hmacKey);
+            sessionController.savePreferences(sessionController.getApplicationPreference());
         }
-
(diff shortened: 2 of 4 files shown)


buddhika75 (commented):
All three issues addressed in commit 4c669418a3:

**1. Timing side-channel (SecurityController:265)** — replaced `String.equals()` with `MessageDigest.isEqual()` on decoded raw bytes. Added `import java.security.MessageDigest`.

**2. Duplicated/racy key-init (NotificationController:286)** — added `obtainHmacSigningKey(SessionController)` to `SecurityController` as the single canonical place to get-or-create the signing key (generates 32-char random key if none exists, persists via `savePreferences`). Both inline blocks in `NotificationController` replaced with a single call to this helper.

**3. Legacy link generators still active (PatientReportController:390)** — migrated all four methods (`emailMessageBody`, `cancelApprovalEmailMessageBody`, `generateQrCodeLink`, `generateQrCodeDetails`) to use HMAC tokens. Removed the dead-code block in `generateQrCodeDetails` that built an unused `temId`. Updated the TODO comment to confirm no new legacy links are generated by this class, making the 2026-07-09 removal date safe.

buddhika75 (commented):
Addressed in commit 50929a59f1:

**SecurityController:273 (compilation error)** — Already fixed in the previous commit (217e625e): removed `NumberFormatException` from the multi-catch since it is a subclass of `IllegalArgumentException`. Only `IllegalArgumentException` remains.

**PatientReportController:1534 (wrong ID used in token)** — Fixed. All four link generators now use `r.getId()` instead of `currentPatientReport.getId()`:
- `emailMessageBody(PatientReport r)`
- `cancelApprovalEmailMessageBody(PatientReport r)`
- `generateQrCodeLink(PatientReport r)`
- `generateQrCodeDetails(PatientReport r)` — inner null guard also changed from `currentPatientReport != null` to `r != null`

PR #20042, opened 2026-04-18
Fix display bug to print 2 barcodes when creating a new sample barcode.

damithdeshan98:
Fix display bug to print 2 barcodes when creating a new sample barcode.

Signed-off-by: damithdeshan98 <hkddrajapaksha@gmail.com>


PasinduW99 (approved):
approved

buddhika75 (approved):

PR #20047, opened 2026-04-18
Fix migration SQL errors in v2.1.3 (SIGNAL in SELECT CASE) and v2.1.4 (UNION ALL on non-existent table)

### LINKED ISSUE #20046: Fix database migration SQL errors in v2.1.3 (SIGNAL in SELECT CASE) and v2.1.4 (UNION ALL on non-existent tables)
## Problem

Two migration scripts fail on the database migration page (`/admin/database_migration.xhtml`):

### v2.1.3 — `SIGNAL` inside `SELECT CASE` (MySQL syntax error 1064)

**Error:**
```
SQLSyntaxErrorException: You have an error in your SQL syntax ... near 'SIGNAL SQLSTATE '45000' SET MESSAGE_TEXT = CONCAT(...)' at line 3
```

**Root cause:** `migration.sql` contains:
```sql
SELECT CASE
    WHEN @null_percentage > 90 THEN
        SIGNAL SQLSTATE '45000'
        SET MESSAGE_TEXT = CONCAT(...)
    ELSE 'Data integrity check passed: Acceptable NULL value levels'
END AS data_integrity_enforcement_check;
```
`SIGNAL` is a statement, not an expression. MySQL's parser rejects it inside a `SELECT CASE...END`. It is only valid as a standalone top-level statement or inside a stored procedure/trigger body.

### v2.1.4 — `UNION ALL` referencing a table that doesn't exist (MySQL error 1146)

**Error:**
```
SQLSyntaxErrorException: Table 'coop.user_stock' doesn't exist
```

**Root cause:** `01-pre-migration-checks.sql` uses:
```sql
SELECT ... FROM USER_STOCK WHERE (subquery checking if table exists) > 0
UNION ALL
SELECT ... FROM userstock WHERE (subquery checking if table exists) > 0
UNION ALL
SELECT 'No table found' WHERE ...
```
MySQL validates **all table references at parse time**, before executing any `WHERE` clause. If only one casing of the table exists (`userstock` on Linux/Ubuntu, `USER_STOCK` on Windows), MySQL throws error 1146 for the missing variant regardless of the runtime `WHERE` guard in `INFORMATION_SCHEMA`.

### v2.1.3
Remove the `SELECT CASE ... SIGNAL ...` enforcement block. The `SELECT ... data_integrity_status` immediately above it already displays the result clearly. The real migration work (the `ALTER TABLE` column changes) has already completed before this check runs.

### v2.1.4
Replace the three-branch `UNION ALL` record-count block with a query that only reads from `INFORMATION_SCHEMA` (always present), avoiding any direct reference to tables that may not exist under one casing.

## Steps to Reproduce

1. Navigate to `/admin/database_migration.xhtml`
2. Attempt to run migration v2.1.3 or v2.1.4
3. Error is thrown and the migration is marked as failed

## Affected Files

- `src/main/resources/db/migrations/v2.1.3/migration.sql`
- `src/main/resources/db/migrations/v2.1.4/01-pre-migration-checks.sql`

buddhika75:
## Summary

- **v2.1.3**: Removes a `SELECT CASE ... SIGNAL ...` block that fails with MySQL error 1064. `SIGNAL` is a statement, not an expression — it cannot appear inside `SELECT CASE...END`. The `data_integrity_status` `SELECT` immediately above already reports CRITICAL/WARNING/SUCCESS, so no separate enforcement is needed.
- **v2.1.4**: Replaces a three-branch `UNION ALL` over `USER_STOCK`/`userstock` with an `INFORMATION_SCHEMA`-only query. MySQL validates all table references at parse time before any `WHERE` guard executes, so referencing a table that exists only under one casing (lowercase on Linux, uppercase on Windows) throws error 1146 unconditionally.

Fixes #20046

## Test plan

- [ ] Navigate to `/admin/database_migration.xhtml`
- [ ] Run migration v2.1.3 — should complete without error 1064
- [ ] Run migration v2.1.4 — should complete without error 1146 on either Linux (lowercase table) or Windows (uppercase table)
- [ ] Verify v2.1.3 data integrity status row still shows CRITICAL/WARNING/SUCCESS as appropriate
- [ ] Verify v2.1.4 pre-checks still report the table variant count correctly via INFORMATION_SCHEMA

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed a faulty conditional enforcement in a migration; migrations now report data-integrity status (CRITICAL/WARNING/SUCCESS) instead of attempting to force failure via query evaluation.

* **Chores**
  * Enhanced pre-migration checks to detect table name variants via schema inspection and provide clearer existence/status reporting for safer migrations.

### DIFF
--- a/src/main/resources/db/migrations/v2.1.3/migration.sql
+++ b/src/main/resources/db/migrations/v2.1.3/migration.sql
@@ -250,7 +250,8 @@ SELECT
     END AS data_integrity_status;
 
 -- NOTE: SIGNAL is a statement, not an expression — it cannot appear inside SELECT CASE...END.
--- Enforcement is handled by the data_integrity_status row above which already flags CRITICAL/WARNING/SUCCESS.
+-- Reporting is handled by the data_integrity_status row above, which flags CRITICAL/WARNING/SUCCESS.
+-- The current migration runner does not fail based on SELECT result rows.
 
 -- ==========================================
 -- MIGRATION COMPLETION WITH FINAL VERIFICATION
--- a/src/main/resources/db/migrations/v2.1.3/migration.sql
+++ b/src/main/resources/db/migrations/v2.1.3/migration.sql
@@ -249,17 +249,8 @@ SELECT
          ELSE 'SUCCESS: Acceptable NULL value levels'
     END AS data_integrity_status;
 
--- ENFORCE SUCCESS: Signal error if data integrity is severely compromised
-SELECT CASE
-    WHEN @null_percentage > 90 THEN
-        SIGNAL SQLSTATE '45000'
-        SET MESSAGE_TEXT = CONCAT(
-            'MIGRATION v2.1.3 DATA INTEGRITY FAILURE: ', @null_percentage,
-            '% NULL values detected in critical financial columns. ',
-            'Data corruption likely occurred. Manual DBA intervention required immediately.'
-        )
-    ELSE 'Data integrity check passed: Acceptable NULL value levels'
-END AS data_integrity_enforcement_check;
+-- NOTE: SIGNAL is a statement, not an expression — it cannot appear inside SELECT CASE...END.
+-- Enforcement is handled by the data_integrity_status row above which already flags CRITICAL/WARNING/SUCCESS.
 
 -- ==========================================
 -- MIGRATION COMPLETION WITH FINAL VERIFICATION
--- a/src/main/resources/db/migrations/v2.1.4/01-pre-migration-checks.sql
+++ b/src/main/resources/db/migrations/v2.1.4/01-pre-migration-checks.sql
@@ -31,22 +31,16 @@ WHERE TABLE_SCHEMA = DATABASE()
 ORDER BY INDEX_NAME;
 
 -- Count total records to estimate impact
-SELECT 'Counting records in USER_STOCK/userstock tables...' AS status;
-
--- Try uppercase table first
-SELECT CONCAT('Total records in USER_STOCK: ', COUNT(*)) AS record_count
-FROM USER_STOCK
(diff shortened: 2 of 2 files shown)


buddhika75:
Addressed CodeRabbit's comment (c8218a8031): reworded the v2.1.3 comment from 'Enforcement is handled by...' to 'Reporting is handled by...' and added a note that the migration runner does not fail based on SELECT result rows.

buddhika75:
Extended the scope to unblock v2.1.0 and v2.1.17, which were also failing on coop production after the v2.1.3/v2.1.4 fix deployed:

- **v2.1.0**: `Duplicate column name 'consumption_allowed'` (1060) — column already exists
- **v2.1.17**: `Can't DROP 'FK_PATIENTSAMPLE_RETIRER_ID'` (1091) — FK never existed under that name on coop

Rather than guarding 46+ DDL statements across multiple files individually, commit 068cfa0d46 generalises the existing `isDuplicateIndexError` handler in `DatabaseMigrationController` into `getIdempotentSkipReason()`, which now silently skips three benign idempotency errors:
- CREATE INDEX already exists or target table missing (1061 / 1146)
- ALTER TABLE ADD COLUMN where the column already exists (1060)
- ALTER TABLE DROP FOREIGN KEY / DROP INDEX on a non-existent key (1091 / 1025)

All three represent "ensure-state" intent where a skip is safe, matching the existing pattern for CREATE INDEX.

buddhika75:
Miss in the earlier fix: the UNION ALL pattern was also in **migration.sql** (lines 45–57), which is the file the runner actually executes. The earlier fix only touched `01-pre-migration-checks.sql`, which turns out to be a standalone pre-check script not invoked by the runner.

Commit 006f8c22e9 applies the same INFORMATION_SCHEMA-only variant count fix to `migration.sql`. After deploy, Re-Execute on v2.1.4 should clear.

PR #20466, opened 2026-05-05
fix(channel): completeBooking NPE — coop production hotfix

### LINKED ISSUE #20461: fix(channel): NullPointerException in completeBooking API causes HTML error response
## Summary

The `/api/channel/complete` endpoint returns an HTML JSF error page instead of a JSON response due to an unhandled `NullPointerException` in `ChannelService.settleOnlineAgentInitialBooking()`.

## Root Cause

In `settleOnlineAgentInitialBooking()`, after `getBillFacade().editAndCommit(paidBill)` is called, the code immediately accesses:

```java
OnlineBooking bookingDetails = paidBill.getReferenceBill().getOnlineBooking();
```

`referenceBill` is declared as `@ManyToOne(fetch = FetchType.LAZY)` on `Bill`. After `editAndCommit()`, the JPA persistence context may detach or re-merge the entity, causing the lazy proxy for `referenceBill` to resolve to `null`. This causes a `NullPointerException` which escapes the JAX-RS layer unhandled, resulting in Payara returning an HTML error page.

Contributing factor: `Bill.copy()` explicitly has `referenceBill` commented out (line 1044), so the only source of truth for `referenceBill` on the completed bill is the explicit `setReferenceBill()` call made inside `saveAgentOnlinePaymentCompletionBill()` — making it sensitive to JPA context state after `editAndCommit()`.

## Fix

Obtain `bookingDetails` directly from `preBillSession.getBill().getOnlineBooking()` (the pending bill, which is already loaded in the persistence context) instead of navigating via `paidBill.getReferenceBill()`. The pending bill IS the reference bill — so the result is identical but without the lazy-load risk.

## Stack Trace

```
java.lang.NullPointerException
    at com.divudi.service.ChannelService.settleOnlineAgentInitialBooking(ChannelService.java:2737)
    at com.divudi.ws.channel.ChannelApi.completeBooking(ChannelApi.java:1460)
```

## Impact

- The `/api/channel/complete` endpoint has **never successfully completed** on the coop deployment — all 7 pending bookings in the database have timed out and been retired
- Affects all online booking agents (doc990 and any future integrations)
- The HTML response confuses API clients expecting JSON

## Steps to Reproduce

1. Call `POST /api/channel/save` with valid session and agency → succeeds (202)
2. Call `POST /api/channel/complete` with the returned `refNo` → returns HTML JSF login/error page

## Verified

Reproduced live on `coop.carecode.org` on 2026-05-05 and confirmed via server log.

buddhika75:
## Summary

Hotfix for `coop-prod` — cherry-picked from PR #20462 (merged to `development`).

The `/api/channel/complete` endpoint was returning an HTML JSF error page instead of JSON for all calls, due to three chained bugs triggered by JPA lazy-loading and JTA flush timing.

### Bugs fixed

1. **Lazy `referenceBill` NPE** (`ChannelService.java`) — after `editAndCommit()`, the `@ManyToOne(LAZY) referenceBill` proxy on `paidBill` became unresolvable. Fixed by reading `onlineBooking` directly from `preBillSession.getBill()` (already in persistence context).

2. **`savePaidBillItem()` returns null** (`ChannelService.java`) — `create()` only calls `persist()` with no flush, so `bi.getId()` is null under JTA. `find(null)` returns null → NPE. Fixed by using `createAndFlush(bi)` and returning `bi` directly.

3. **`bookingData` null after re-fetch** (`ChannelApi.java`) — after settlement the `OnlineBooking` was re-fetched by ref no, but the auto-retire background timer had already set `RETIRED=1`. Fixed by using the pre-fetched `bookingData` reference for patient/booking fields and deriving financial totals from `completedBill.getNetTotal() + agencyCharge`.

## Impact

- `/api/channel/complete` has never successfully completed on coop production
- All 7 pending bookings in the DB timed out and were retired due to this bug
- Affects doc990 integration (and any future online booking agent)

## Testing

- Reproduced and fixed locally against coop DB backup
- Full end-to-end test (save → complete) confirmed returning `202` JSON with `status: ACTIVE`
- Also verified against staging (`stg.carecode.org/coop`) which uses the production DB

Closes #20461

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/src/main/java/com/divudi/service/ChannelService.java
+++ b/src/main/java/com/divudi/service/ChannelService.java
@@ -2809,8 +2809,8 @@ public class ChannelService {
         bi.copy(bs.getBillItem());
         bi.setCreatedAt(new Date());
         bi.setBill(b);
-        getBillItemFacade().create(bi);
-        return getBillItemFacade().find(bi.getId());
+        getBillItemFacade().createAndFlush(bi);
+        return bi;
     }
 
     private void savePaidBillFee(Bill b, BillItem bi, BillSession bs) {
--- a/src/main/java/com/divudi/service/ChannelService.java
+++ b/src/main/java/com/divudi/service/ChannelService.java
@@ -2757,7 +2757,7 @@ public class ChannelService {
 
         List<Payment> p = createPayment(paidBill, paidBill.getPaymentMethod());
 
-        OnlineBooking bookingDetails = paidBill.getReferenceBill().getOnlineBooking();
+        OnlineBooking bookingDetails = preBillSession.getBill().getOnlineBooking();
         bookingDetails.setOnlineBookingPayment(agencyCharge);
         bookingDetails.setHospitalFee(paidBill.getHospitalFee());
         bookingDetails.setDoctorFee(paidBill.getStaffFee());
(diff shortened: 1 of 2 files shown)


damithdeshan98 (approved):
Approved

PR #20493, opened 2026-05-06
Add fee-level service charge allowed toggle for inward billing fees

buddhika75:
### Motivation
- Provide a per-fee switch to opt out of inward service charge (margin) calculation so base fees can independently disable margins while preserving existing item-level `marginNotAllowed` behavior. 
- Preserve default behavior where margins are calculated unless explicitly disabled at the fee level.

### Description
- Added `private boolean marginAllowed = true;` to `Fee` with `isMarginAllowed()`/`setMarginAllowed()` accessors so new fees default to allowing margins. 
- Updated inward margin logic in `InwardBeanController.setBillFeeMargin(...)` (both overloads) to skip margin calculation when `billFee.getFee()` is `null` or `!billFee.getFee().isMarginAllowed()`. 
- Exposed the toggle in the Base Fees UI at `/admin/pricing/manage_item_fees.xhtml` in the add/edit dialog (`itemFeeManager.itemFee.marginAllowed`) and in the fees table (`f.marginAllowed`) with labels "Service Charge Allowed / Service Charge NOT Allowed". 
- Files changed: `src/main/java/com/divudi/core/entity/Fee.java`, `src/main/java/com/divudi/bean/inward/InwardBeanController.java`, and `src/main/webapp/admin/pricing/manage_item_fees.xhtml`.

### Testing
- No automated build or test suite was executed for this change (per repository policy no Maven/compile/test run unless explicitly requested).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69fbbd588db0832fa0341c046568bdbd)

### DIFF
--- a/src/main/java/com/divudi/bean/inward/InwardBeanController.java
+++ b/src/main/java/com/divudi/bean/inward/InwardBeanController.java
@@ -2382,7 +2382,9 @@ public class InwardBeanController implements Serializable {
     public void setBillFeeMargin(BillFee billFee, Item item, PriceMatrix priceMatrix) {
         double margin = 0;
 
-        if (billFee == null || item.isMarginNotAllowed()) {
+        if (billFee == null || item.isMarginNotAllowed()
+                || billFee.getFee() == null
+                || !billFee.getFee().isMarginAllowed()) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


buddhika75:
Closing in favour of #20494 which fixes a critical defect in this PR: the primitive `boolean marginAllowed = true` field would cause EclipseLink to add the DB column as `NOT NULL DEFAULT 0`, setting all existing fees to service-charge-not-allowed on deploy and breaking live inward billing. #20494 uses a nullable `Boolean` with a `Boolean.FALSE.equals()` guard so existing rows (NULL) are treated as allowed, preserving current behaviour.

PR #20984, opened 2026-05-23
hotfix(lims): skip wrong-component rows when writing analyzer results (#20981)

### LINKED ISSUE #20981: LIMS: Analyzer results overwrite wrong component in multi-sample investigations (BSP, OGTT)
## Problem

When an analyzer sends results for a multi-sample investigation (e.g. Blood Sugar Profile with BSP/PB, BSP/PL, BSP/PD time points, or OGTT with FBS/1Hr/2Hr), the middleware incorrectly overwrites earlier component values instead of placing each result in its correct time-point slot.

### Root Cause

In `LimsMiddlewareController.addResultToReport()`, the value-writing loop iterates all `PatientReportItemValue` rows. For each row whose test code matches the incoming result, it:

1. **Path 0** (lines ~1135–1164): tries to match `patientSampleComponant.investigationComponant == priv.sampleComponent` — works correctly when both are set
2. **Fallback** (line ~1166): fires when `valueToSave == false`  
   - Branch 1: writes if either side is null (correct — single-component tests)
   - Branch 2: writes if `priv.sampleComponent == ps.investigationComponant` (correct)
   - **Branch 3 (else)**: writes unconditionally even when both sides are set but *don't match* — this overwrites the wrong component's row

### Observed Effect

Confirmed from coop production data (PatientReport 9577263 for BLOOD SUGAR PROFILES (3 SAMPLE), pi 9562840):
- BSP/PL result (117.2) was written to the BSP/PB row instead of only its own BSP/PL row
- Both BSP/PB and BSP/PL end up with the same value; each new result overwrites all previous time points

### Affected Investigations

Any investigation with `HASMORETHANONECOMPONANT = 1` and analyzer-driven result entry:
- BLOOD SUGAR PROFILE (id 87603)
- BLOOD SUGAR PROFILES (3 SAMPLE) (id 87604)
- OGTT - (3 SAMPLE) (id 87644)

### Safe Path (Unaffected)

Single-component tests (FBS, FBC, etc.) where `ps.investigationComponant` is null always take Branch 1 — they are unaffected by this fix.

## Fix

Remove the erroneous write from the `else` branch in `addResultToReport()`. When both `ps.investigationComponant` and `priv.sampleComponent` are non-null but don't match, the row belongs to a different time point and must be skipped — the correct matching row will be handled in a later iteration of the same loop.

The same `else` in `addResultToReportPrevious()` was already a no-op (just a comment) — no change needed there.

buddhika75:
## Summary

Hotfix cherry-picked from PR #20982 (merged to `development`).

- Analyzer results for multi-sample investigations (BSP, OGTT) were overwriting earlier time-point values instead of placing each result in its correct slot
- Root cause: the `else` branch in `addResultToReport()` wrote unconditionally to any `PatientReportItemValue` whose test code matched, even when `priv.sampleComponent` and `ps.investigationComponant` were both set but belonged to **different** time points
- Fix: `else` branch is now a no-op — mismatched component rows are skipped; the correct row is handled in the next iteration

## Test plan

- [x] Tested all scenarios on local instance against coop data — confirmed working by developer before requesting hotfix

## Safety

Single-component tests (`ps.investigationComponant == null`) are unaffected — they always take Branch 1.

Fixes #20981

### DIFF
--- a/src/main/java/com/divudi/ws/lims/LimsMiddlewareController.java
+++ b/src/main/java/com/divudi/ws/lims/LimsMiddlewareController.java
@@ -1208,26 +1208,9 @@ public class LimsMiddlewareController {
                                     temFlag = true;
                                     valueToSave = true;
                                 } else {
-                                    // // System.out.println("3 result = " + result);
-                                    priv.setStrValue(result);
-
-                                    Double dbl = 0d;
-                                    try {
-                                        dbl = Double.parseDouble(result);
-                                        // // System.out.println("3 dbl = " + dbl);
-                                    } catch (Exception e) {
-                                    }
-                                    priv.setDoubleValue(dbl);
-                                    // // System.out.println("3 priv.getDoubleValue() = " + priv.getDoubleValue());
-                                    if (priv.getId() == null) {
-                                        // // System.out.println("3 new priv created = " + dbl);
-                                        patientReportItemValueFacade.create(priv);
-                                    } else {
-                                        // // System.out.println("3 new priv Updates = " + dbl);
-                                        patientReportItemValueFacade.edit(priv);
-                                    }
-                                    temFlag = true;
-                                    valueToSave = true;
+                                    // Both ps.investigationComponant and priv.sampleComponent are set
+                                    // but they don't match — this row belongs to a different time point.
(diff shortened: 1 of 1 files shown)


damithdeshan98 (approved):
Approved

Issue #21025, opened 2026-05-26
Counsumtion bug


Irani96:
File Path: reports/inventoryReports/consumption.xhtml

- [ ] Incorrect Return Quantity and Amount Displayed in Consumption Report

bill numbers : R/DIS/RHD/LBK/26/000079  / DIS/ML/RHD/26/000141

                        
Description:

Returned item quantity and values are incorrectly displayed in the Consumption Report.

Item Details:

Item: TRIMETHOPRIM-SULFAMETHOXAZOLE 1.25 / 23.75 µg (CT0052)
Actual Return Quantity: 8

Issue:

Consumption report displays quantity as 16 instead of 8
Return value section displays quantity 6 only, but amount is calculated for quantity 8 (Rs. 11,200)
Displayed values are inconsistent with the actual returned quantity

Displayed Values:

Purchase Value: -8,400.00
Cost Value: -8,400.00
Retail Value: -8,400.00

Expected Behavior:
Consumption report should display the correct return quantity (8)

<img width="1359" height="632" alt="Image" src="https://github.com/user-attachments/assets/15f46fd7-115e-44f7-8b85-eb78a629ec15" />

- [ ] Purchase, Cost, and Retail Values Displayed as 0 in Consumption Report

Description:
In the Consumption Report, the following values are displayed as 0:

Purchase Value
Cost Value
Retail Value

Only the Net Value is displayed correctly.

<img width="1366" height="677" alt="Image" src="https://github.com/user-attachments/assets/1d827de5-6dfe-47ff-95a0-895c87ef9f5a" />

Bill number : DIS/LBK/RHD/26/000249   (IMMERSION OIL - 10 ml    1)
R/DIS/RHD/LBK/26/000111   DIS/ML/RHD/26/000161 IMMERSION OIL - 10 ml   -1)
DIS/LBK/RHD/26/000249  (IMMERSION OIL - 10 ml  1)

PR #21031, opened 2026-05-27
fix(pharmacy-report): show net consumption in Disposal Consumption Report (#21025) [coop hotfix]

buddhika75:
## Summary

Hotfix to `coop-prod` mirroring the fix in #21030 (against `development`).

Issue [#21025](https://github.com/hmislk/hmis/issues/21025) reported the Disposal Consumption Report inflating quantity and showing 0 for Purchase / Cost / Retail Value while Net Value looked correct. Root cause: `billItem.qty` is stored positive on both issue and return bills, while `billItemFinanceDetails.valueAt*Rate` / `billFinanceDetails.total*Value` use a stock-direction sign (negative on stock-out, positive on stock-back-in). Aggregations summed the two conventions raw, so qty over-counted and value columns cancelled.

This change normalises each row to consumption direction (issue +, return / cancel −) before aggregating; `netTotal` already carries the correct sign and is left alone. The byBill / byBillItem tables now expose per-row consumption-direction values on `PharmacyRow` so the on-screen rows and the footer total agree, and the PDF export reads the same fields.

Data is uniformly in the canonical convention (verified against ruhunu prod, 12.8 K issue rows / 160 return rows all in the expected sign). **No data backfill required** — the fix is purely reporting logic.

## Test plan
- [ ] Open Consumption Report on coop-prod, run as Summary across a date range that includes both issues and returns; confirm Purchase / Cost / Retail aggregate to net consumption (not ~0).
- [ ] By Bill Item filtered to one item with returns: return rows display negative qty and value, footer matches row sum.
- [ ] By Bill: same at bill granularity.
- [ ] PDF / Excel byBill export values match the on-screen rows.

Refs #21025
Refs #21030

### DIFF
--- a/src/main/java/com/divudi/bean/pharmacy/PharmacyController.java
+++ b/src/main/java/com/divudi/bean/pharmacy/PharmacyController.java
@@ -3857,6 +3857,50 @@ public class PharmacyController implements Serializable {
         }
     }
 
+    /**
+     * Multiplier to convert a stored {@code valueAt*Rate} into consumption-direction value.
+     *
+     * <p>Stored {@code billItemFinanceDetails.valueAt*Rate} (and the matching
+     * {@code billFinanceDetails.total*Value}) use a stock-direction sign: negative when
+     * stock leaves the source department, positive when it comes back. Consumption
+     * accounting is the inverse — an issue adds to consumption, a return/cancel subtracts.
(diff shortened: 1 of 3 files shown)


buddhika75:
Closing — hotfix should target ruhunu-prod, not coop-prod. Replaced by a new PR targeting ruhunu-prod.

PR #21058, opened 2026-05-28
fix(migration) : v2.9.0 restores AUTO_INCREMENT on ID primary keys

### LINKED ISSUE #21028: Error page appreared when adding the Inward Services & Investigation
**Path**
Inward → Service & Investigation → Add Service & Investigation
inward/inward_bill_service.xhtml

**Steps**
1. Select an Item (ex: X-Ray ) → Click on the 'Add' button
2. Click on the settle button

**Expected Result**
User should be able to view bill

Actual Result
Error page appreared when adding the Inward Services & Investigation

Image

damithdeshan98:
## Summary

  Some databases (observed on **Roseth**) were restored from dumps that dropped the `AUTO_INCREMENT` attribute on `ID`
  primary-key columns. The entities map `@GeneratedValue(strategy = IDENTITY)`, so EclipseLink omits `ID` on `INSERT`
  and reads it back via `SELECT LAST_INSERT_ID()` — which only works when the column is `AUTO_INCREMENT`. Without it,
  every insert fails with:

  ```
  java.sql.SQLException: Field 'ID' doesn't have a default value  (error 1364)
  ```

  This broke inward bill settlement (`inward/inward_bill_service.xhtml`) on Settle. This is a **database-side** issue,
  not an application code bug, so it is repaired through a versioned migration.

  ## Changes

  - **New migration `v2.9.0`** — scans `INFORMATION_SCHEMA` for every table whose single-column `BIGINT` `ID` primary
  key lacks `AUTO_INCREMENT` and re-adds it (preserving the existing column type) via a temporary stored procedure with
  a cursor.
    - **Idempotent** — a no-op on healthy databases (scan returns no rows). Safe to re-run.
    - **Safe** — disables FK checks and relaxes `sql_mode` only for the duration of the rebuild (so child-table ALTERs
  succeed and legacy invalid defaults such as `TIMESTAMP DEFAULT '0000-00-00 00:00:00'` do not trigger error 1067), then
   restores both session settings.
    - **Non-destructive** — no row data changes; MySQL auto-sets each table's next value to `MAX(id)+1`.
    - **Cross-deployment safe** — detects the actual stored case of the `ID` column per table, so it works on both
  case-sensitive (Linux) and case-insensitive (Windows) MySQL.
    - Includes BEFORE/AFTER verification `SELECT`s in the script.
    - `rollback.sql` is a documented **no-op** (removing `AUTO_INCREMENT` again would re-break inserts).

  - **`MigrationInfo.java`** — `verificationQueries` retyped from `List<String>` to `Object` so a migration-info file
  whose `verificationQueries` is an object grouping queries by environment (e.g. `v2.1.5`'s `{"common": [...],
  "production": [...]}`) no longer throws `MismatchedInputException` and aborts the migration-list load.
  `hasVerificationQueries()` updated to handle `Collection`, `Map`, and scalar shapes.

  ## Files

  - `src/main/resources/db/migrations/v2.9.0/migration.sql`
  - `src/main/resources/db/migrations/v2.9.0/migration-info.json`
  - `src/main/resources/db/migrations/v2.9.0/rollback.sql`
  - `src/main/java/com/divudi/core/util/MigrationInfo.java`

  ## Prerequisites

  - Database backup completed before running.
  - Datasource user has the `CREATE ROUTINE` privilege (the migration creates and drops a temporary stored procedure).

  ## Test plan

  - [ ] Run migration `v2.9.0` from the admin *Database Migration* page on an affected DB.
  - [ ] BEFORE report lists the affected tables; AFTER report returns **no rows**.
  - [ ] Inward bill settlement at `inward/inward_bill_service.xhtml` succeeds with no error 1364.
  - [ ] Re-running the migration on an already-fixed DB is a clean no-op.
  - [ ] Migration list loads without `MismatchedInputException` for `v2.1.5`.

  Closes #21028

## Summary by CodeRabbit

* **Chores**
  * Added a v2.9.0 database migration to restore AUTO_INCREMENT on ID primary keys; requires a maintenance window and may take ~30 minutes.
  * Improved migration metadata handling to accept verification queries in multiple formats (array or grouped object) so migration checks are more flexible.

[Review Change Stack](https://app.coderabbit.ai/change-stack/hmislk/hmis/pull/21058?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/main/java/com/divudi/core/util/MigrationInfo.java
+++ b/src/main/java/com/divudi/core/util/MigrationInfo.java
@@ -4,8 +4,10 @@
  */
 package com.divudi.core.util;
 
+import java.util.Collection;
 import java.util.Date;
 import java.util.List;
+import java.util.Map;
 
 /**
  * Data class for migration metadata parsed from migration-info.json
@@ -28,7 +30,10 @@ public class MigrationInfo {
     private List<String> postMigrationVerification;
     private List<String> rollbackNotes;
     private List<MigrationStep> migrationSteps;
-    private List<String> verificationQueries;
+    // Tolerates either a JSON array (["q1", "q2"]) or an object grouping queries by
+    // environment ({"common": [...], "production": [...]}). Typed as Object so a single
+    // file's shape variation never fails the whole migration list load.
+    private Object verificationQueries;
 
     // Inner class for migration steps
     public static class MigrationStep {
@@ -205,11 +210,11 @@ public class MigrationInfo {
         this.migrationSteps = migrationSteps;
     }
 
-    public List<String> getVerificationQueries() {
+    public Object getVerificationQueries() {
         return verificationQueries;
     }
 
-    public void setVerificationQueries(List<String> verificationQueries) {
+    public void setVerificationQueries(Object verificationQueries) {
         this.verificationQueries = verificationQueries;
     }
 
@@ -223,7 +228,16 @@ public class MigrationInfo {
     }
 
     public boolean hasVerificationQueries() {
-        return verificationQueries != null && !verificationQueries.isEmpty();
+        if (verificationQueries == null) {
+            return false;
+        }
+        if (verificationQueries instanceof Collection) {
+            return !((Collection<?>) verificationQueries).isEmpty();
+        }
+        if (verificationQueries instanceof Map) {
+            return !((Map<?, ?>) verificationQueries).isEmpty();
+        }
+        return true;
     }
 
     public boolean hasMigrationSteps() {
--- a/src/main/resources/db/migrations/v2.9.0/migration-info.json
+++ b/src/main/resources/db/migrations/v2.9.0/migration-info.json
@@ -1,0 +1,18 @@
+{
+  "version": "v2.9.0",
+  "description": "Restore AUTO_INCREMENT on every ID primary key that lost it during a database import",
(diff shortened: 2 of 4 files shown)


buddhika75 (changes requested):
1. migration.sql silently swallows ALTER TABLE failures at lines 71-73. The migration runner drains SELECT results but does not fail the migration if the AFTER query still returns rows, so this can be recorded as SUCCESS even when  some tables remain broken. That should be changed to collect/log failures and SIGNAL at the end, or simply let the error fail the migration.
  2. migration-info.json says requires_downtime: false, but this alters primary-key columns across potentially large tables (bill, billitem, archives, etc.). These ALTER TABLE ... MODIFY COLUMN operations can take metadata/table locks and block writes. This should be treated as a maintenance-window migration, or at least the PR should clearly state operational risk.

damithdeshan98:
> 1. migration.sql silently swallows ALTER TABLE failures at lines 71-73. The migration runner drains SELECT results but does not fail the migration if the AFTER query still returns rows, so this can be recorded as SUCCESS even when  some tables remain broken. That should be changed to collect/log failures and SIGNAL at the end, or simply let the error fail the migration.
> 2. migration-info.json says requires_downtime: false, but this alters primary-key columns across potentially large tables (bill, billitem, archives, etc.). These ALTER TABLE ... MODIFY COLUMN operations can take metadata/table locks and block writes. This should be treated as a maintenance-window migration, or at least the PR should clearly state operational risk.

It's now Fixed.

buddhika75 (approved):

Issue #20932, opened 2026-05-21
Need to improve Route Analysis Report Print, PDF, Excel


Irani96:
File Path: reports/collectionCenterReports/route_analysis_report.xhtml

- [ ] After clicking the Hide Chart , the chart is still displayed in the print- Report Type(Summary,Detail)
- [ ] After clicking the Hide Chart , the chart is still displayed in the excel- Report Type(Summary,Detail)
- [ ] After clicking the Hide Chart , the chart is still displayed in the pdf- Report Type(Summary,Detail)

RashmikaHarshamal:
Done, Should be merge changes

chathushkiupulini:
path cannot be reproduced because the system displays an error page before reaching the expected screen. As a result, the functionality cannot be tested.

<img width="1903" height="956" alt="Image" src="https://github.com/user-attachments/assets/33d9f3c1-f8c7-428d-bd74-599a2530d082" />

RashmikaHarshamal:
Check with qa2. it have updated

PR #21710, opened 2026-06-27
#20932

RashmikaHarshamal:
## Summary by CodeRabbit

* **Bug Fixes**
  * Route analysis exports now include charts only when chart display is enabled, preventing unwanted chart pages or images in Excel and PDF outputs.
  * The route analysis report view now respects the chart visibility setting, so the count chart is hidden when chart display is turned off.


RashmikaHarshamal:
Done

buddhika75 (approved):

PR #21770, opened 2026-07-01
fix(pharmacy): guard NPE in saveItemBatchWithCosting when rates are null

### LINKED ISSUE #20947: GRN Costing Save & Approve Error for if only add free qty
File Path: pharmacy/pharmacy_grn_costing_with_save_approve.xhtml

Title: GRN Costing Save & Approve Throws NullPointerException

Description:
An error page is displayed when clicking the “Save & Approve” button in GRN Costing. The transaction fails with a java.lang.NullPointerException.

Affected Page:
/qa4/faces/pharmacy/pharmacy_grn_costing_with_save_approve.xhtml

Root Cause:
Null value encountered in:

GrnCostingController.java:4343
Method: saveItemBatchWithCosting()

buddhika75:
## Summary
- `saveItemBatchWithCosting()` crashed with NPE when a GRN line had only free qty and no purchase rate
- `lineNetRate`, `retailSaleRatePerUnit`, `totalCostRate`, and `unitsPerPack` on `BillItemFinanceDetails` can all be null
- Wrapped all four reads with `BigDecimalUtil.valueOrZero()` so a zero-rate item batch is created instead of crashing

## Test plan
- [ ] GRN with free-qty-only line, no purchase rate — Save and Approve completes without error
- [ ] Normal GRN lines with qty and rates continue to work

Closes #20947

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved item costing calculations to handle missing or zero values more safely.
  * Added fallback logic to calculate per-unit retail rates from pack-level rates when needed.
  * Made rate and unit conversions more resilient, helping reduce pricing and costing errors during stock entry.


buddhika75:
@coderabbitai review

buddhika75:
Fixed in second commit `7615246e4d`: when `retailSaleRatePerUnit` is null or zero, we now fall back to `retailSaleRate / unitsPerPack` (pack-level retail rate divided by units per pack). This preserves the actual retail price from the GRN line rather than silently storing zero on the ItemBatch.

PR #21771, opened 2026-07-01
fix(inward): disable Add Service button on click to prevent double-submit

### LINKED ISSUE #21576: Add Service Button Can Be Clicked Multiple Times
File Path: /inward/inward_timed_service_consume.xhtml

Description:
The Add Service button remains active after being clicked, allowing users to click it multiple times. This can result in duplicate service entries being added unintentionally.

Steps to Reproduce:

Open the service entry screen
Select a service
Click the Add Service button multiple times rapidly

Actual Result:
The Add Service button can be clicked multiple times

Expected Behavior:
The Add Service button should be disabled immediately after the first click until the request is completed

buddhika75:
## Summary
- The Add Service button on `inward_timed_service_consume.xhtml` had no double-click protection
- Rapid clicks before the non-AJAX page reload could create duplicate timed service entries
- Added `onclick="this.disabled=true;"` to disable immediately on first click
- Added stable `id="btnAddTimedService"` for accessibility

## Test plan
- [ ] Click Add Service rapidly — only one entry should be created
- [ ] Single click still works; button re-enables after page reload

Closes #21576

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented accidental form submissions when pressing Enter, except when interacting with a textarea.
  * Improved the “Add Service” action to reduce duplicate requests by adding a client-side guard that blocks rapid repeated clicks for a short period.

### DIFF
--- a/src/main/webapp/inward/inward_timed_service_consume.xhtml
+++ b/src/main/webapp/inward/inward_timed_service_consume.xhtml
@@ -10,7 +10,7 @@
 
 
     <ui:define name="content">
-        <h:form>    
+        <h:form id="formTimedServiceConsume" onkeypress="if(event.keyCode == 13 &amp;&amp; event.target.type != 'textarea') { return false; }">
             <h:panelGroup class="row" rendered="#{inwardTimedItemController.current.patientEncounter eq null}"> 
                 <div class="col-12">
                     <h:panelGroup>
--- a/src/main/webapp/inward/inward_timed_service_consume.xhtml
+++ b/src/main/webapp/inward/inward_timed_service_consume.xhtml
@@ -245,7 +245,7 @@
                                             class="mx-2 ui-button-success"
                                             value="Add Service"
                                             ajax="false"
-                                            onclick="this.disabled=true;"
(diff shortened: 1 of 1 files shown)


buddhika75:
@coderabbitai review

buddhika75:
Fixed in the second commit `af05c7df5c`: replaced `onclick="this.disabled=true;"` with a flag-based guard — `if(this.alreadyClicked) return false; this.alreadyClicked=true; setTimeout(...reset..., 5000);`

The button is never disabled, so its client ID is always present in the POST and PrimeFaces can decode the action correctly. A 5-second timeout resets the flag for error recovery.

buddhika75:
Fixed in commit `0758f1b7f7`: added the project-standard Enter-key guard `onkeypress="if(event.keyCode == 13 &amp;&amp; event.target.type != 'textarea') { return false; }"` to the `h:form` element, consistent with the pattern used in `inward_assessment_discharge_medicines.xhtml`, `ward_pharmacy_return_to_pharmacy.xhtml`, and other forms. Also added a stable `id="formTimedServiceConsume"` to the form.

PR #21773, opened 2026-07-01
fix(inward): lazy-load encounterCreditCompanys in BHT edit view

### LINKED ISSUE #21244: Credit Company Details Not Displayed in Ward Admission
File Path: inward/inward_edit_bht.xhtml

Description:
Credit company details entered during ward admission are not displayed in the relevant screens/reports.

Affected Fields:

Credit Company
Reference Number

Actual Result:

Credit Company and Reference Number are not displayed

Expected Behavior:

Credit Company and Reference Number entered during ward admission should be displayed correctly

buddhika75:
## Summary
- `getEncounterCreditCompanys()` returned null when the user navigated directly to the BHT edit page without going through a navigation action that called `fillCreditCompaniesByPatient()`
- The credit company panel then showed no rows even though `EncounterCreditCompany` records existed in the DB
- Added null guard in the getter: if the list is null and a current encounter is set, `fillCreditCompaniesByPatient()` is called on demand

## Test plan
- [ ] Navigate directly to BHT edit for a patient with credit company records
- [ ] Credit Company panel should display the records without needing any extra action
- [ ] Adding a new credit company still works correctly

Closes #21244

### DIFF
--- a/src/main/java/com/divudi/bean/inward/BhtEditController.java
+++ b/src/main/java/com/divudi/bean/inward/BhtEditController.java
@@ -977,6 +977,9 @@ public class BhtEditController implements Serializable, ControllerWithPatient {
     }
 
     public List<EncounterCreditCompany> getEncounterCreditCompanys() {
+        if (encounterCreditCompanys == null && current != null) {
+            fillCreditCompaniesByPatient();
+        }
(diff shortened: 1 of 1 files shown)


buddhika75:
@coderabbitai review

PR #21781, opened 2026-07-01
fix: search families by any member's NIC or phone number

thisarasamuditha:
Family search previously matched only the Chief Householder's NIC and the Family's own phone number, so lookups by a non-CHH member's NIC or phone silently returned no results even though the member was registered.

- searchFamily(): also match FamilyMember.patient.person.phone/mobile via EXISTS, in addition to Family.phoneNo and membershipCardNo
- searchFamilyByNIC(): also match FamilyMember.patient.person.nic via EXISTS, in addition to the CHH's NIC
- searchFamilyMember(): add missing NIC condition alongside the existing phone/card matching

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved family and membership search so results can now be found using additional linked patient details, including phone, mobile, and NIC.
  * Enhanced family lookup matching to include related non-retired family members, making searches more complete and accurate.

### DIFF
--- a/src/main/java/com/divudi/bean/common/PatientController.java
+++ b/src/main/java/com/divudi/bean/common/PatientController.java
@@ -927,7 +927,7 @@ public class PatientController implements Serializable, ControllerWithPatient {
         quickSearchPhoneNumber = null;
         admissionController.setPatientAllergies(null);
         admissionController.setCurrentReservation(null);
-        
+
         admissionController.setPatientForiegner(false);
         return "/inward/inward_admission?faces-redirect=true";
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/divudi/bean/common/PatientController.java
```diff
@@ -2423,7 +2423,12 @@ public String searchFamily() {
         String jpql = "Select f "
                 + "from Family f "
                 + "where f.retired=false "
-                + "and (f.phoneNo = :pn or f.membershipCardNo = :mcn) ";
+                + "and (f.phoneNo = :pn "
```
buddhika75, line 2426: We do not use inner JPQL usually. We have to fix this by some other means. Do not use this update. Delete the branch and assign the issue to me.

buddhika75 (commented):
We do not use inner JPQL usually. We have to fix this by some other means. Do not use this update. Delete the branch and assign the issue to me.

Issue #21814, opened 2026-07-02
pharmacy_batches/create: missing stockId in response + suspected long-held lock


buddhika75:
## Context

Found while running a batch stock-verification adjustment against COOP production via `/api/pharmacy_batches/create` + `/api/pharmacy_adjustments/stock_quantity` (same pattern used for the earlier RMH stock verification).

## 1. Confirmed bug — response omits `stockId` when the field is null at serialization time

`PharmacyBatchApiService.createBatch()` (`src/main/java/com/divudi/service/pharmacy/PharmacyBatchApiService.java`) builds the `BatchCreateResponseDTO` immediately after `stockFacade.create(stock)`:

```java
Stock stock = createStockEntry(itemBatch, department);
...
return new BatchCreateResponseDTO(
        itemBatch.getId(),
        stock.getId(),   // <-- can be null if not yet flushed
        ...
);
```

If the JPA provider hasn't assigned/flushed the `Stock` id at that point, `stock.getId()` returns `null`. The API's `Gson` instance (`PharmacyBatchApi.java`) is not configured with `.serializeNulls()`, so the `stockId` key is silently **omitted** from the JSON response rather than being `null` — any client reading `data.stockId` gets a missing-field error instead of a usable value (or an explicit, debuggable `null`). The underlying `Stock` row *is* created correctly at qty=0 — only the response is broken.

Likely the same risk applies to `itemBatch.getId()` used just above it, and `stock.getId()`/`itemBatch.getId()` in similar create-response DTOs elsewhere.

**Suggested fix**: flush before reading the id — e.g. `stockFacade.edit(stock)` (and equivalent for `itemBatch`) right after `create()`, or an explicit `entityManager.flush()` — so the response always contains the real generated id.

## 2. Suspected — long-held lock/transaction on `/pharmacy_batches/create`

During the same session, one call to this endpoint appeared to block for several minutes. Multiple unrelated `POST /pharmacy_adjustments/stock_quantity` calls issued afterward (different `stockId`s, different items) queued up and only resolved sequentially, each completing right after the previous one — consistent with all of them waiting behind whatever the first `/pharmacy_batches/create` transaction was holding. All writes eventually succeeded correctly (verified via the search API afterward — no data loss or double-application), but the delay is far outside normal request latency (a plain `stock_quantity` adjustment completes in well under a second).

[...]

PR #21816, opened 2026-07-02
fix(pharmacy): flush before reading generated id in batch-create API

buddhika75:
## Summary

- `PharmacyBatchApiService.createBatch()` was reading `itemBatch.getId()` and `stock.getId()` immediately after `itemBatchFacade.create(...)`/`stockFacade.create(...)`, which only calls `entityManager.persist(entity)` — no flush.
- Both `ItemBatch` and `Stock` use `@GeneratedValue(strategy = GenerationType.IDENTITY)`, and with `eclipselink.jdbc.batch-writing=JDBC` enabled, the generated id isn't guaranteed to be populated at that point, so `stockId`/`batchId` could come back `null` under load.
- Since `PharmacyBatchApi`'s `Gson` instance wasn't configured with `.serializeNulls()`, a null `stockId` was silently **omitted** from the JSON response instead of appearing as an explicit `null` — callers got a missing-field error instead of a debuggable signal.

## Fix

- Switched both create calls to the existing `AbstractFacade.createAndFlush()` helper (`persist()` + `flush()`, entity stays managed) so ids are guaranteed populated before the response DTO is built.
- Added `.serializeNulls()` to `PharmacyBatchApi`'s `GsonBuilder` as defense-in-depth, so any future unexpected null is visible in the response.

## Scope

This addresses **part 1** of #21814 (the confirmed missing-`stockId` bug). Part 2 (suspected long-held lock on this endpoint) needs production-level concurrency/DB introspection (`SHOW PROCESSLIST` / `SHOW ENGINE INNODB STATUS`) to confirm and is out of scope here.

A background codebase search found the same `create()` → immediate `getId()` gap in 10 other REST API create endpoints (inward, clinical, common, pharmacy-discount, favourite-medicine APIs). Filing a follow-up issue to apply the same fix there rather than bundling it into this PR.

## Test plan

Verified locally (department 485 "Main Pharmacy", AMP 109754 "Leflox 250mg tablet"):
- [x] Single `POST /api/pharmacy_batches/create` call returns non-null `stockId`/`batchId`
- [x] 10 concurrent calls all returned non-null `stockId`/`batchId` (exercises the batch-writing timing window that caused the original intermittent null)
- [x] DB verification: every returned `stockId`/`batchId` matches a real row in `STOCK`/`ITEMBATCH`
- [x] Existing-batch reuse path (same AMP/batchNo/expiry, different department) still returns the correct existing `batchId`, a new `stockId` for the new department, and `message = "Used existing batch"`

Example response after fix:
```json
{"code":200,"data":{"batchId":13859524,"stockId":13859553,"batchNo":"TEST-21814-1", ...},"status":"success"}
```

## Summary by CodeRabbit

* **New Features**
  * API responses now include fields with `null` values instead of omitting them.

* **Bug Fixes**
  * Batch creation now returns newly created record IDs more reliably right away, improving response consistency.

### DIFF
--- a/src/main/java/com/divudi/service/pharmacy/PharmacyBatchApiService.java
+++ b/src/main/java/com/divudi/service/pharmacy/PharmacyBatchApiService.java
@@ -319,7 +319,7 @@ public class PharmacyBatchApiService implements Serializable {
             ib.setWholesaleRate(wholesaleRate);
         }
 
-        itemBatchFacade.create(ib);
+        itemBatchFacade.createAndFlush(ib); // flush so getId() below is guaranteed populated
         return ib;
     }
 
@@ -328,7 +328,7 @@ public class PharmacyBatchApiService implements Serializable {
         stock.setItemBatch(itemBatch);
         stock.setDepartment(department);
         stock.setStock(0.0); // Always start with 0 quantity
-        stockFacade.create(stock);
+        stockFacade.createAndFlush(stock); // flush so getId() below is guaranteed populated
         return stock;
     }
 
--- a/src/main/java/com/divudi/ws/pharmacy/PharmacyBatchApi.java
+++ b/src/main/java/com/divudi/ws/pharmacy/PharmacyBatchApi.java
@@ -56,6 +56,9 @@ public class PharmacyBatchApi {
     private PharmacyBatchApiService batchService;
 
     private static final Gson gson = new GsonBuilder()
+            // Serialize nulls explicitly so an unexpected null field is visible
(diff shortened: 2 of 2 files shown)


buddhika75:
Addressed the outside-diff-range finding from the CodeRabbit review (shared Gson's `.serializeNulls()` affecting `searchOrCreateAmp`/`searchAmp` responses) in follow-up PR #21818, since this PR had already merged before the review completed. Split into two Gson instances so null-serialization is scoped to the batch-create response only; verified locally that the two AMP-search endpoints are unaffected.

PR #21849, opened 2026-07-04
Add config-gated cross-department PO receiving for RMH Hambantota

### LINKED ISSUE #21848: Allow cross-department PO receiving (config-gated) for RMH Hambantota
## Problem

Purchase Orders (POs) are created in one department and, on every current institution, must be received (GRN created) from that same department. The receive-list query filters strictly by the user's currently active department.

RMH Hambantota needs a different workflow: POs are created in the **Pharmacy** department but finalized/approved and received (GRN created) in the **Store** department. Today, `pharmacy_purchase_order_list_for_recieve.xhtml` never shows Pharmacy-created POs to a user working in the Store department, so RMH cannot complete their receiving workflow.

This must be **opt-in per hospital**, defaulting to today's behavior (department must match), so no other institution's workflow changes.

Customer's own words:
> They would like to view POs in a different department. POs have been created in the pharmacy department and are complete in the store department at Hambantota RMH. (When creating POs in Pharmacy, they need to be finalized and approved in the Store department.)
>
> Purchase Orders are created here: `pharmacy/pharmacy_purhcase_order_request.xhtml`
> POs are listed to create GRNs here: `pharmacy/pharmacy_purchase_order_list_for_recieve.xhtml`
>
> Currently, all institutions create the PO in a department, and when the user goes to create a GRN, the listed POs are limited to those created in the same department. RMH wants to create POs in one department and GRN in another. This should be a config option, default OFF, that RMH turns on.

## Current Behavior (traced in code)

- **PO creation** — `PurchaseOrderRequestController.java` (~line 806-808) sets both `bill.department` and `bill.fromDepartment` to the creating user's current department at creation time.
- **GRN receive-list query** — `SearchController.createPoTable(List, List)` and `SearchController.createPoTablePharmacyDto()` both filter with `AND b.department = :dept`, where `:dept` is `sessionController.getDepartment()` — the department the logged-in user is currently working in. This is the hard restriction causing the issue.
- **GRN creation** — `GrnController.createGrn()` sets the new GRN bill's department from the *current session context*, not copied from the PO. So a GRN created while working in Store already lands stock in Store correctly today — there is no business-logic blocker here, only the listing query needs to change.
- **Pages affected by the two query methods above**:
  - `pharmacy_purchase_order_list_for_recieve.xhtml` (main page named in the request)
  - `pharmacy_purchase_order_list_for_recieve_with_approval.xhtml`
  - `pharmacy_wholesale/pharmacy_purchase_order_list_for_recieve.xhtml`
  - `pharmacy_purchase_order_list_for_recieve_dto.xhtml` (DTO variant)
- **Out of scope**: `SearchController.createPoTableStore(...)`, which backs Store's own separate PO/GRN flow (external distributor purchases into Store) — unrelated to this request.

## Design

Add one boolean config option:

```
Pharmacy - Allow Cross-Department PO Receiving   (default: false)
```

Read via the existing generic pattern: `configOptionApplicationController.getBooleanValueByKey(key, false)`.

In `SearchController.createPoTable(List<BillTypeAtomic>, List<BillTypeAtomic>)` and `SearchController.createPoTablePharmacyDto()`, change the JPQL builder so the `AND b.department = :dept` clause (and the `dept` parameter binding) is only appended when the config is `false`. The `AND b.referenceBill.institution = :ins` clause remains unconditional in both cases — this keeps the relaxation scoped to *within one institution* (one hospital's deployment/database) and never leaks POs across institutions.

No new entity, table, or admin UI is needed — this reuses the existing generic config-option mechanism. No changes needed to GRN creation, since it already assigns the GRN to the user's current department regardless of the PO's department.

### Access control

The existing `GoodsRecipt` privilege check on the page (`webUserController.hasPrivilege('GoodsRecipt')`) remains the only gate. A user still needs that privilege while working in whatever department they're currently in to see or receive any PO — cross-department or not. No new privilege is introduced.

### Alternatives Considered

**Explicit department-pair mapping** (e.g. a configurable "Pharmacy → Store" pairing via a new entity/admin UI) was considered but rejected: it's more precise but adds real complexity (new entity, migration, admin screen) for a need that today is a single hospital, single department pair. The simple institution-wide toggle can be revisited if a second hospital needs a narrower, asymmetric rule.

## Testing Plan

- Config `false` (default): behavior unchanged on all four affected pages — POs from other departments remain excluded.
- Config `true`: a PO created in Pharmacy department appears on the receive-list page while the logged-in user's active department is Store (and vice versa), and a GRN can be created against it, landing stock in the currently active department.
- Cross-institution isolation still holds regardless of the config value — a PO belonging to a different institution never appears.

Full design spec: `developer_docs/feature/cross-department-po-receiving.md`

buddhika75:
## Summary
- Adds a new boolean config option `Pharmacy - Allow Cross-Department PO Receiving` (default `false`) via `ConfigOptionApplicationController.loadPharmacyConfigurationDefaults()`.
- When enabled, `SearchController.createPoTable(...)` and `SearchController.createPoTablePharmacyDto()` drop the `b.department = :dept` filter on the PO receive-list queries, while keeping `b.referenceBill.institution = :ins` — so the relaxation is scoped to *within one hospital*, never leaking POs across institutions.
- Fixes #21848: RMH Hambantota creates POs in the Pharmacy department but needs to finalize/receive them (create the GRN) in the Store department. Every other hospital keeps today's same-department-only behavior unless they opt in.
- Full design rationale in `developer_docs/feature/cross-department-po-receiving.md`.

## Test plan
- [x] Config `false` (default): Store department could not see a PO created in Main Pharmacy — "No purchase orders found" (unchanged behavior).
- [x] Config `true`: same search from Store now shows the Main Pharmacy PO.
- [x] Created and finalized a GRN against that PO while logged into Store — GRN bill correctly persisted with `department = Store`, confirmed in the local DB.
- [x] `mvn clean package` — build succeeds, all 117 existing tests pass (including the pre-existing `SearchControllerDepartmentFilterRegressionTest`, unaffected since it targets an unrelated filter).
- [x] Verified against local `coop` DB / Galle Co-Operative Hospital instance; config reverted back to `false` after testing to restore default state.

### Screenshots (also published to the wiki: [Cross-Department PO Receiving](https://github.com/hmislk/hmis/wiki/Cross-Department-PO-Receiving))

**Config off (default) — Store cannot see the Pharmacy PO:**
Store config off

**Config on — Store now sees the Pharmacy PO:**
Store config on

**GRN saved under the Store department:**
GRN saved in Store

## Summary by CodeRabbit

* **New Features**
  * Added a new pharmacy configuration option to allow PO receiving from a different department within the same institution.
  * Updated PO receiving behavior to respect this setting when showing eligible records.

* **Documentation**
  * Expanded configuration and workflow docs with the new pharmacy procurement option and department-scoping details.
  * Clarified the limits of cross-department receiving and how it affects receiving screens.


buddhika75:
Re: the Codex review comment about scoping this toggle by institution — investigated this. \`getBooleanValueByKey(...)\` is intentionally global (\`OptionScope.APPLICATION\`, no institution) for every one of the ~150 existing config options in \`ConfigOptionApplicationController\` — there's no institution-scoped boolean getter/setter anywhere in the codebase today. Adding one just for this option would be inconsistent with every other toggle and is out of scope for this change.

In practice, each production hospital (coop, ruhunu, southernlanka, etc.) runs its own separate database deployment, so a global toggle is scoped by deployment in real use. The local \`coop\` test DB happens to also host a lightly-used secondary institution ("Co op Diagnostic Centre") which would technically share this flag if enabled there, but that's a property of the shared config mechanism as a whole, not something specific to this PR — leaving as-is to stay consistent with existing conventions.

PR #21854, opened 2026-07-04
fix(pharmacy): F15 report adjustment values never reconciled (BillFinanceDetails never populated)

buddhika75:
## Summary

The F15 "Daily Stock Values" pharmacy report's Adjustment Transactions section always showed 0.00, even on days with real pharmacy adjustment activity — breaking the reconciliation identity `Opening + Sales + Purchases + Transfers + Adjustments = Closing`. Root cause: the `/api/pharmacy_adjustments/*` REST endpoints create audit-trail bills but never populate `BillFinanceDetails` or `Bill.total`/`netTotal`, which is exactly what the F15 report's adjustment query reads. Reproduced live against COOP production (OPD Pharmacy, 03 July 2026) via Playwright before starting the fix.

- Added the missing `PHARMACY_STOCK_EXPIRY_DATE_AJUSTMENT` bill type to the report's adjustment query list
- Fixed stock-quantity adjustments to record the delta value `(afterQty − beforeQty) × rate`, not the full ending value, and populate `BillFinanceDetails` + bill totals
- Fixed retail-rate and purchase-rate adjustments to populate `BillFinanceDetails` + bill totals (their delta formula was already correct)
- Fixed expiry-date adjustments to create an explicit zero-valued `BillFinanceDetails` for consistency
- Added an idempotent, privilege-gated admin backfill endpoint (`POST /pharmacy_adjustments/backfill_finance_details`, dry-run by default) to retroactively populate `BillFinanceDetails` for adjustment bills created before this fix existed, using only each bill's own stored audit fields — never inventing data

Full design/implementation plan: `docs/superpowers/plans/2026-07-04-f15-adjustment-value-fix.md`

## Test plan
- [x] `mvn test` — full suite passes
- [x] Each task implemented via TDD with dedicated unit tests (dummy-facade convention, no Mockito)
- [x] Task-level and final whole-branch code review completed; all Critical/Important findings fixed (delta-vs-ending-value semantics, cost-value approximation disclosure, before/after totals, privilege scope widened to cover all 3 adjustment bill types the backfill touches, backfill `toDate` end-of-day normalization, missing `@Transactional` boundary)
- [ ] Run the admin backfill in dry-run mode against COOP OPD Pharmacy production data, verify output, then apply and confirm F15 reconciles for 2026-07-03 and 2026-07-04 (tracked separately — requires a `-hotfix` branch to `coop-prod` per repo convention, not part of this PR)

## Summary by CodeRabbit

* **New Features**
  * Added an admin-only endpoint to backfill missing pharmacy historical finance details and bill totals, with dry-run support (`apply=false`) and department/date-range targeting.
  * Pharmacy adjustment capabilities now include expiry-date adjustments.
* **Bug Fixes**
  * Pharmacy adjustment bills now immediately populate finance details and update bill totals correctly when created.
  * Backfill operation now skips bills that already have finance details and improves historical valuation accuracy by preferring stored snapshots for retail and cost rates.

### DIFF
--- a/src/main/java/com/divudi/core/data/dto/adjustment/BackfillRequestDTO.java
+++ b/src/main/java/com/divudi/core/data/dto/adjustment/BackfillRequestDTO.java
@@ -1,0 +1,17 @@
+package com.divudi.core.data.dto.adjustment;
+
+public class BackfillRequestDTO {
+    private Long departmentId;
+    private String fromDate;
+    private String toDate;
+    private boolean apply;
+
+    public Long getDepartmentId() { return departmentId; }
+    public void setDepartmentId(Long departmentId) { this.departmentId = departmentId; }
+    public String getFromDate() { return fromDate; }
+    public void setFromDate(String fromDate) { this.fromDate = fromDate; }
+    public String getToDate() { return toDate; }
+    public void setToDate(String toDate) { this.toDate = toDate; }
+    public boolean isApply() { return apply; }
+    public void setApply(boolean apply) { this.apply = apply; }
+}
--- a/src/main/java/com/divudi/core/data/dto/adjustment/BackfillResultDTO.java
+++ b/src/main/java/com/divudi/core/data/dto/adjustment/BackfillResultDTO.java
@@ -1,0 +1,28 @@
+package com.divudi.core.data.dto.adjustment;
+
+public class BackfillResultDTO {
+    private Long billId;
+    private String billTypeAtomic;
+    private double computedNetTotal;
+    private double computedTotal;
+    private boolean applied;
+    private String note;
+
+    public Long getBillId() { return billId; }
+    public void setBillId(Long billId) { this.billId = billId; }
+
+    public String getBillTypeAtomic() { return billTypeAtomic; }
+    public void setBillTypeAtomic(String billTypeAtomic) { this.billTypeAtomic = billTypeAtomic; }
+
+    public double getComputedNetTotal() { return computedNetTotal; }
+    public void setComputedNetTotal(double computedNetTotal) { this.computedNetTotal = computedNetTotal; }
+
+    public double getComputedTotal() { return computedTotal; }
+    public void setComputedTotal(double computedTotal) { this.computedTotal = computedTotal; }
+
+    public boolean isApplied() { return applied; }
+    public void setApplied(boolean applied) { this.applied = applied; }
+
+    public String getNote() { return note; }
+    public void setNote(String note) { this.note = note; }
+}
--- a/src/main/java/com/divudi/core/entity/Bill.java
+++ b/src/main/java/com/divudi/core/entity/Bill.java
@@ -3065,6 +3065,16 @@ public class Bill implements Serializable, RetirableEntity {
         return billFinanceDetails;
     }
 
+    /**
+     * Null-check for billFinanceDetails that does NOT auto-vivify a new instance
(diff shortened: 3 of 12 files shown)


buddhika75:
Tested this branch against a full local restore of the COOP production database (F15 report for 2026-07-04 / OPD Pharmacy) and found two additional bugs that only surfaced with real data, now fixed in commits 280273f199 and d1fefca08b:

1. **Backfill date-range truncation**: `billFacade.findByJpql(jpql, params)` binds `Date` params as `TemporalType.DATE`, silently dropping the `endOfDay()` time component and excluding same-day bills from the backfill. Fixed by binding `from`/`to` explicitly as `TIMESTAMP`.
2. **Missing purchase-value delta for stock-quantity adjustments**: `PHARMACY_STOCK_ADJUSTMENT` bills only computed cost/retail deltas, leaving `totalPurchaseValue` at zero even though a quantity change moves stock value at all three rates. Fixed in both the live and backfill paths, without double-counting into the bill's own `netTotal`.
3. **Cost/Retail valuation used current rate instead of historical snapshot**: `StockHistoryFacade.calculateStockValueAtRetailRateOptimized`/`calculateStockValueAtCostRateOptimized` always read the item batch's *current* rate, unlike the already-correct purchase-rate query which prefers the `STOCKHISTORY` row's own rate snapshot. This made a mid-period `PHARMACY_RETAIL_RATE_ADJUSTMENT` invisible to Opening/Closing valuation, double-counting the adjustment bill's delta in the report's reconciliation. Fixed to mirror the purchase-rate query.

After all three fixes, Cost, Purchase, and Retail Rate columns of the F15 report reconcile exactly (Opening + Sales + Purchases + Transfers + Adjustments = Closing) for the tested date/department. Full test suite passes.

buddhika75:
Continued testing against the local COOP production restore for 2026-07-03 (OPD Pharmacy) surfaced two more issues, now fixed in commit 4a686bce82:

1. **Duplicate/retry bills in the source data** (not a code bug — a data issue from the original bulk stock-verification session): item batch 13860953 had three `PHARMACY_STOCK_ADJUSTMENT` bills at the identical timestamp, all claiming the same `before=0, after=400` correction. Since `resetStock()` does an absolute set, only one had a real effect, but summing all three triple-counted it. Retired the two redundant bills in the local test copy; production needs the same check (query + steps now in `tmp/coop/2026JulyOpdPharmacy/COOP_Production_Fix_TODO.md`, tracked in project memory for the eventual production pass — not part of this PR's diff since it's a data change, not code).

2. **`backfillFinanceDetails` used the item's current cost/purchase rate instead of the real historical rate snapshot.** `PharmacyBean.addToStockHistory` already records a cost/purchase/retail rate snapshot at the exact time of every stock movement (the same mechanism retail already used via `BillItem.netRate`) — this just wasn't applied to cost/purchase, on the mistaken assumption no such snapshot existed. Added `StockHistoryFacade.findByPharmaceuticalBillItem()` and now prefer that snapshot, falling back to current rate (with disclosure) only when none exists. This is what actually explained the remaining Purchase Rate column gap (7,374.40 for 2026-07-03) after retiring the duplicate bills above.

After both fixes, Cost, Purchase, and Retail Rate columns all reconcile exactly for 2026-07-03, OPD Pharmacy, in addition to the previously-verified 2026-07-04. Full test suite passes, including new regression coverage for the historical-rate-snapshot lookup.

PR #22041, opened 2026-07-11
17867 create report for outside payments

### LINKED ISSUE #17867: Create report for Outside Payments
File Path: reports/financialReports/outside_payment.xhtml

Data file Path: inward/inward_bill_outside_charge.xhtml
Memo can be type
Paid Amount can be type

https://ruhunuhospitals-my.sharepoint.com/:x:/g/personal/admin_ruhunuhospital_lk/EeEGSuwLOcNFu01_PtQZ1ggBnFB6mbBmNlQctYwd_yn2Jw?e=XU581C&nav=MTVfezAwMDAwMDAwLTAwMDEtMDAwMC0wQzAwLTAwMDAwMDAwMDAwMH0

thisarasamuditha:
## Summary by CodeRabbit

* **New Features**
  * Added an Outside Payment Report with filters for admission/discharge/invoice-approval date ranges, institution/site/department, referring doctor, credit company, admission type, and paid/not-paid status.
  * Added a results table showing invoice/patient details, charge info, paid amount, due amount, and cancellation/refund indicators.
  * Enabled inline updates for paid status, paid amount, and memo with confirmation.
  * Added print, Excel, and PDF export options, including validation and user notifications when no records match.


src/main/java/com/divudi/bean/inward/OutsideChargeReportController.java
```diff
@@ -0,0 +1,511 @@
...
+            boolean paid = row.getPaid() != null && row.getPaid();
+            bill.setPaid(paid);
+            bill.setPaidAt(paid ? new Date() : null);
+            billFacade.edit(bill);
+
+            if (row.getBillItemId() != null) {
+                BillItem billItem = billItemFacade.find(row.getBillItemId());
+                if (billItem != null) {
+                    billItem.setDescreption(row.getMemo());
+                    billItemFacade.edit(billItem);
```
buddhika75, line 254: **Payment date is silently overwritten on every save.** `bill.setPaidAt(paid ? new Date() : null)` runs unconditionally whenever `updateOutsidePaymentRow` is called. If a bill was already paid weeks ago (its `paidAt` set then), and the user later edits only the Memo field for that row and clicks Update, the checkbox is still `true`, so `paidAt` is reset to `now` — the original payment date is lost. Consider only setting `paidAt` when transitioning from unpaid→paid (e.g. `if (paid && bill.getPaidAt() == null) bill.setPaidAt(new Date());` and clear it on paid→unpaid).

src/main/java/com/divudi/bean/inward/OutsideChargeReportController.java
```diff
@@ -0,0 +1,511 @@
...
+
+        if (invoiceApprovedFromDate != null && invoiceApprovedToDate != null) {
+            jpql.append("AND b.createdAt BETWEEN :iafd AND :iatd ");
+            params.put("iafd", invoiceApprovedFromDate);
+            params.put("iatd", invoiceApprovedToDate);
+        }
+
+        if (institution != null) {
+            jpql.append("AND b.institution = :ins ");
+            params.put("ins", institution);
```
buddhika75, line 167: **"Invoice Approved" date filter queries the wrong column.** The filter labelled *Invoice Approved From/To Date* appends `AND b.createdAt BETWEEN :iafd AND :iatd`, i.e. it filters on bill *creation* time, not approval. `Bill` has dedicated `approveAt` and `invoiceDate` fields — one of those is almost certainly what this filter is meant to use. As written, results and the Excel metadata will be misleading for any deployment where invoices are approved well after creation.

src/main/java/com/divudi/bean/inward/OutsideChargeReportController.java
```diff
@@ -0,0 +1,511 @@
...
+
+        if (dischargeFromDate != null && dischargeToDate != null) {
+            jpql.append("AND enc.dateOfDischarge BETWEEN :dfd AND :dtd ");
+            params.put("dfd", dischargeFromDate);
+            params.put("dtd", dischargeToDate);
+        }
+
+        if (invoiceApprovedFromDate != null && invoiceApprovedToDate != null) {
+            jpql.append("AND b.createdAt BETWEEN :iafd AND :iatd ");
+            params.put("iafd", invoiceApprovedFromDate);
```
buddhika75: **Discharge / Invoice-Approved "To" dates are not end-of-day normalized, so the last day is excluded.** `dischargeToDate` / `invoiceApprovedToDate` come straight from a date-only `p:datePicker` (time = 00:00:00) and are used in `BETWEEN :dfd AND :dtd` with `TemporalType.TIMESTAMP`. Every discharge/approval that happened after midnight on the selected end day is dropped. Note `fromDate`/`toDate` getters already call `getStartOfDay`/`getEndOfDay`; the discharge and invoice-approved getters (lines ~427-457) return the raw field with no such normalization. Apply `CommonFunctions.getStartOfDay(...)` / `getEndOfDay(...)` to these bounds before building the query.

src/main/java/com/divudi/bean/inward/OutsideChargeReportController.java
```diff
@@ -0,0 +1,511 @@
...
+                + "COALESCE(enc.bhtNo, ''), "
+                + "COALESCE(item.name, ''), "
+                + "enc.dateOfDischarge, "
+                + "b.cancelled, "
+                + "b.refunded, "
+                + "bi.inwardChargeType, "
+                + "COALESCE(createrPerson.name, ''), "
+                + "bi.createdAt, "
+                + "bi.netValue, "
+                + "b.paidAt, "
```
buddhika75, line 123: **`enc.patient.phn` and `enc.patient.person.name` are implicit INNER joins in a `SELECT NEW` — rows with a null patient/person are silently dropped, and this two-/three-hop path off a join alias has been observed to silently return zero rows in EclipseLink (see project notes on #23222 and the SELECT-NEW nested-LEFT-JOIN gotcha).** Replace with explicit `LEFT JOIN enc.patient pt LEFT JOIN pt.person pr` and select `COALESCE(pt.phn, )`, `COALESCE(pr.name, )`.

src/main/java/com/divudi/bean/inward/OutsideChargeReportController.java
```diff
@@ -0,0 +1,511 @@
...
+                return;
+            }
+            bill.setPaidAmount(row.getPaidAmount() != null ? row.getPaidAmount() : 0.0);
+            boolean paid = row.getPaid() != null && row.getPaid();
+            bill.setPaid(paid);
+            bill.setPaidAt(paid ? new Date() : null);
+            billFacade.edit(bill);
+
+            if (row.getBillItemId() != null) {
+                BillItem billItem = billItemFacade.find(row.getBillItemId());
```
buddhika75, line 251: **`updateOutsidePaymentRow` leaves `bill.paidAmount` inconsistent when the row is marked *not* paid.** It always does `bill.setPaidAmount(row.getPaidAmount() ...)` before checking the paid flag, so unchecking *Paid* while a stale amount sits in the input persists `paidAmount>0, paid=false, paidAt=null`. When `paid` is false the paid amount should be forced to 0 (or left untouched only deliberately).

src/main/java/com/divudi/bean/inward/OutsideChargeReportController.java
```diff
@@ -0,0 +1,511 @@
...
+
+        workbook.setSheetName(0, "Outside Payment Report");
+        sheet.shiftRows(0, sheet.getLastRowNum(), 7);
+
+        Map<String, Object> filters = getFiltersForOutsidePaymentReport();
+        if (filters != null && !filters.isEmpty()) {
+            pharmacyController.addMetaDataToExcelSheet(workbook, sheet, 0, "Outside Payment Report", filters);
+        }
+    }
+
```
buddhika75, line 317: **Excel metadata layout only *just* fits the 7 rows reserved by `shiftRows(0, lastRowNum, 7)`.** `addMetaDataToExcelSheet` writes: institution row + title row + ceil(N/3) filter rows. With the 13 filters built here that is 2 + 5 = 7 rows (0-6), exactly the gap opened by the shift. Adding a 16th filter later will silently overwrite the exported table header with no error. Prefer computing the shift amount from the actual metadata height (or have `addMetaDataToExcelSheet` return the used row count and shift by that).

src/main/java/com/divudi/bean/inward/OutsideChargeReportController.java
```diff
@@ -0,0 +1,511 @@
...
+            }
+
+            JsfUtil.addSuccessMessage("Row updated successfully.");
+        } catch (Exception e) {
+            Logger.getLogger(OutsideChargeReportController.class.getName())
+                    .log(Level.SEVERE, "Error updating outside payment row", e);
+            JsfUtil.addErrorMessage("Error updating row.");
+        }
+    }
+
```
buddhika75, line 272: **After `updateOutsidePaymentRow`, `update="tblobs"` re-renders the table from the same `reportRows` list whose DTO date fields are `final` and never refreshed.** The *Paid Date* column will keep showing the old value (blank) even though the bill now has `paidAt = now`, and *Created Date* etc. are also stale relative to the DB. Re-run `processOutsidePaymentReport()` (or patch the DTO in place) at the end of the update so the grid reflects what was saved.

src/main/java/com/divudi/core/data/dto/OutsidePaymentReportDto.java
```diff
@@ -0,0 +1,181 @@
...
+    public void setPaidAmount(Double paidAmount) {
+        this.paidAmount = paidAmount;
+    }
+
+    public Double getInvoiceTotal() {
+        return invoiceTotal;
+    }
+
+    public Double getDueAmount() {
+        return invoiceTotal - (paidAmount != null ? paidAmount : 0.0);
```
buddhika75, line 153: **[merge-gate code review]** `getDueAmount()` = `invoiceTotal - paidAmount` returns a large-negative value for cancelled / refunded bills, because the JPQL selects `b.netTotal` into `invoiceTotal` and cancellation/refund bills carry a negative `netTotal`. That negative Due Amount is then rendered in the report grid and PDF (`pdfNumCell(row.getDueAmount())`) right next to live rows, and summed into any total. Consider zeroing or flagging the amount columns when `cancelled`/`refunded` is true, the same way `getCancelledOrRefundedLabel()` already special-cases those rows for display.

buddhika75:
## 🚦 Merge Gate: BLOCKED — blocking code-review findings

**Phase 1 (high-effort code review):** 8 findings posted inline (discussion_r3910572113, r3910573309, r3910573399, r3910573493, r3910573597, r3910575529, r3910575776, r3910582060). **4 are High-severity correctness / business-rule issues:**

1. **`updateOutsidePaymentRow` clobbers the original payment date** — `bill.setPaidAt(paid ? new Date() : null)` runs on *every* row save, so editing an unrelated field (Memo) on an already-paid row overwrites its true `paidAt` with "now". Should only stamp on an unpaid→paid transition.
2. **"Invoice Approved From/To Date" filter queries the wrong column** — it filters `b.createdAt` instead of an approval timestamp (`Bill` has `approveAt` / `invoiceDate`). Results are wrong wherever approval lags creation.
3. **Discharge / Invoice "To Date" silently excludes the last selected day** — those pickers are date-only (time 00:00:00) and feed a `BETWEEN … TIMESTAMP`; unlike `fromDate`/`toDate`, their getters don't normalize via `getEndOfDay`.
4. **`SELECT NEW` uses implicit INNER joins on `enc.patient.phn` and `enc.patient.person.name`** — rows with a null patient/person are silently dropped, and this multi-hop path off a join alias has previously returned zero rows in EclipseLink (project note on #23222). Needs explicit `LEFT JOIN` + `COALESCE`.

Plus 3 Medium (un-checking "Paid" persists `paidAmount>0 / paid=false`; Excel `shiftRows(..., 7)` hardcodes room for exactly the current 13-filter metadata block; the grid re-renders from a `reportRows` list whose `final` DTO date fields are never refreshed after an update, so "Paid Date" stays blank) and 1 Low (`getDueAmount()` goes large-negative for cancelled/refunded bills and is still rendered + summed).

The DTO 19-arg `SELECT NEW` mapping, the `creater`/`webUserPerson` LEFT JOIN, `Doctor extends Staff`, the `p:dataExporter` editable-cell handling, and `BillType.InwardOutSideBill` were all checked and are fine. The branch also compiles cleanly and the report page renders without error (Process returned 0 rows only because this local DB has no `InwardOutSideBill` data).

Merge-gate stopped here; Phase 2/3 (build already done, E2E, baselines) did not run. Also note the branch is ~2 months behind `development` and its green CI checks are from 2026-07-11 — a rebase + fresh CI is advisable before merge regardless.

buddhika75:
## 🚦 Merge Gate: fixes applied (commit 0f5b5470e5)

Addressed the blocking review findings from the previous run:

| # | Fix |
|---|---|
| 1 | `updateOutsidePaymentRow` now stamps `paidAt` **only on an unpaid → paid transition**. Editing the memo (or anything else) on an already-paid row no longer overwrites the original payment date. Un-marking as paid clears it. |
| 4 | JPQL reaches the patient/person via explicit `LEFT JOIN enc.patient pt` / `LEFT JOIN pt.person ptPerson` instead of the dotted `enc.patient.person.name` inside `SELECT NEW` (no more implicit INNER-join row drops / EclipseLink multi-hop zero-rows risk). |
| 3 | Discharge & Invoice-Approved "To Date" bounds are widened with `getStartOfDay`/`getEndOfDay`, so a `BETWEEN … TIMESTAMP` no longer excludes everything on the last selected day. |
| 5 | `paidAmount` is forced to 0 when the row is not paid — un-checking "Paid" can't leave `paidAmount>0 / paid=false`. |
| 6 | Excel `shiftRows` offset is derived from the real metadata height (`1 + 1 + ceil(filters/3) + 1`) instead of a hardcoded `7` (which was already one row short for the current 13-filter block). |
| 7 | After a successful bill edit, `paidAmount`/`paid` are patched back onto the row DTO so the re-rendered grid and the derived Due Amount are correct without a full reprocess. (Paid Date stays a reprocess-only refresh — that DTO field is `final`; noted in a comment.) |
| 8 | `getDueAmount()` returns 0 for cancelled/refunded rows instead of a large-negative value. |

**Finding 2 (Invoice Approved filter keys on `b.createdAt`)** — kept as-is. That is the same proxy the sibling inpatient reports use (`InwardReportController` → `pe.finalBill.createdAt`), and outside-charge bills never populate `approveAt`, so switching would make the filter match nothing. Documented in a code comment.

**Verified live** against the Galle coop DB with two seeded `InwardOutSideBill` rows:
- Report returns both rows; charge-type labels ("Scanning Charges", "Laboratory Charges"), creater name and patient name all resolve through the new joins.
- Mark row paid → `paidAt` stamped once (`10:14:03`).
- Edit the memo on that now-paid row → `paidAt` **preserved** at `10:14:03` (server time was `10:14:44`).
- Un-check Paid → `paid=0`, `paidAmount=0`, `paidAt=NULL`; grid Due Amount goes back to the full 3,500.00 without reprocessing.

Compiles clean. Also worth noting the branch is ~2 months behind `development` and its green CI is from 2026-07-11 — a rebase + fresh CI is still advisable before merge. Re-run merge-gate once CI is green for a fresh verdict.

Issue #21918, opened 2026-07-06
Itemized Service Summary Report shows Incorrect Data on Bill Cancellation


shinyZu:
**Describe the bug**

- The **Itemized Service Summary - OPD and Inward (Fast - DTO-Based)** report does not accurately reflect cancelled, returned, or refunded service bills.
- Additionally, the service fee is doubled in the report, resulting in incorrect financial information.

**Report URL:** /inward/inward_itemized_service_summary_dto.xhtml
- Inward Analytics → Service Reports → Itemized Service Summary - OPD and Inward (Fast - DTO-Based)

**To Reproduce**

1. Login to the **Inward** department.
2. Admit a patient (or use an existing admitted patient).
3. Navigate to **Add Services and Investigations**.
4. Add a service (e.g., MRI - ABDOMEN) to the admitted patient.
5. Use relevant filters, generate and review the **Itemized Service Summary** - OPD and Inward (Fast - DTO-Based) report.
6. **Cancel** the service bill that was created.
7. Generate the report again and compare the results.
8. Observe that:
       - Cancelled service bills are not reflected correctly.
       -  The service fee is displayed as double after the cancellation.
9. Repeat the same steps using the Refund and Return processes (if applicable).
10. Generate the report after each transaction and observe that the report data is still inaccurate.

**Expected behavior**
- The report should accurately reflect the status of cancelled, returned, and refunded service bills.
- Service fees should be displayed with the correct amount.
- Report totals should match the actual billing transactions.

**Screenshots**

- Screenshot for Step 5 - Before Cancellation
<img width="1920" height="986" alt="Image" src="https://github.com/user-attachments/assets/52bd1b05-d0f3-4134-98c6-bd96d4a2e254" />
</br></br>

- Screenshot for Step 7 - After Cancellation
<img width="1917" height="980" alt="Image" src="https://github.com/user-attachments/assets/55c587e1-4ace-40e3-bbfe-3c96e7708e5a" />

buddhika75:
## Verified: already fixed on `development`

Investigated this issue and found the underlying bug was already fixed as a side effect of PR #22051 (closing #21920, merged 2026-07-12) and refined in PR #22053 (closing #22050). Neither PR explicitly closed this issue, so it stayed open even though the fix landed.

[...]

PR #22051, opened 2026-07-12
Show per-billing rows with billed date in Itemized Service Summary (#21920)

### LINKED ISSUE #21920: Itemized Service Summary Report Incorrectly Updates Same Record on Re-billing
**Description**

- This issue is related to: [Itemized Service Summary Report shows Incorrect Data on Bill Cancellation
 ]( https://github.com/hmislk/hmis/issues/21918)

- In addition to the previously reported [issue]( https://github.com/hmislk/hmis/issues/21918), the **Itemized Service Summary - OPD and Inward (Fast - DTO-Based)** report has the following additional issues:

1. When the same service (e.g., MRI - ABDOMEN) is billed again after cancellation/refund, the report does not create a new record. Instead, it updates the existing record.
2. As a result, every time the same service is re-billed for inpatients, **the same record gets overwritten** instead of maintaining separate entries.
3. The report also **does not include a “Billed Date” column**, making it difficult to track when each service was actually billed.

**To Reproduce**

1. Login to Inward department
2. **Add** a service to an admitted patient via **Add Services and Investigations** ( Example: MRI - ABDOMEN)
3. Generate the Itemized Service Summary report
5. **Cancel/refund** the service bill
6. **Re-add** the same service again for the same patient
7. Generate the report again
8. Observe that:
   - The same record is updated instead of creating a new entry
   - Historical billing entries are not preserved
9. Repeat billing the same service multiple times and observe that the same record keeps getting overwritten
10. Try the same service with different patients.

**Expected behavior**

- Each billing instance of a service should be recorded as a separate row in the report
- Previously billed services should not be overwritten
- The report should include a **“Billed Date” column** to track when each service was billed
- Users should be able to see complete billing history per service

**Actual behavior**

- Re-billed services overwrite existing records in the report
- No separate tracking of repeated billing events
- Missing "Billed Date" column

**Screenshots**

- Step 3 - Report before Cancellation
Image
</br></br>

-  Report after Cancellation
Image
</br></br>

- Step 7 - Report after Re-adding the same service
Image

buddhika75:
## Summary

Fixes the **Itemized Service Summary - OPD and Inward (Fast - DTO-Based)** report (`/inward/inward_itemized_service_summary_dto.xhtml`), which grouped every billing of a service into a single row. Re-billing, cancelling or refunding the same service overwrote the existing record, and there was no way to see when each service was billed.

Closes #21920 (builds on the cancellation/fee concerns raised in #21918).

## What changed

- New `BillService.fetchItemizedServiceInstanceDTOs(...)` returns **one row per `BillItem`** (no group-by), so every billing/cancellation/refund is preserved as its own row.
- Added **Billed Date**, **BHT** and **Patient** columns; `rowKey` is now the `BillItem` id (previously `itemId_staffName`, which collided across instances).
- Cancellation/refund bill items are **already stored negative**, so their fee/value columns are used as-is — only the row **count** gets a synthesized sign. This shows cancellations as separate negative rows and avoids the fee-doubling that the aggregated report exhibits (#21918).
- Used explicit **LEFT JOINs** for `patientEncounter`/`patient` so OPD rows (no encounter) aren't dropped by an implicit inner join.
- `OpdSaleSummaryDTO` gains 4 fields + one **new** constructor (existing constructors untouched, per the DTO guidelines).

Scope is isolated: the OPD-only *Itemized Sales Summary* page (`/opd/analytics/itemized_sale_summary_dto.xhtml`) still uses the original aggregated `fetchOpdSaleSummaryDTOs` and is unchanged. No schema change (no new persisted fields), so no DDL regeneration needed.

## Verification (local Payara + Playwright + DB cross-check)

Exercised on the local `coop` DB against real data.

**Per-instance rows + Billed Date + Patient** — filtering one service (MRI - ABDOMEN) now returns **22 separate rows** instead of a single aggregated row:

Per-instance rows

**Cancellation is a separate negative row (not overwritten)** — for BHT/53223 (patient G K DANAWATHI), FBC billed at 07:25:54 (**+456.00**) and cancelled at 07:56:03 (**Count −1, Net −456.00**) appear as two separate rows. Grand Total (2,000.00) matches a direct `SUM(bi.netValue)` over the same window:

Cancellation separate negative row

**DB cross-check** confirmed the report rows (dates, BHT, patient, signs) match the underlying `billitem`/`bill` data exactly. The OPD *Itemized Sales Summary* page was regression-checked and still aggregates as before.

## Summary by CodeRabbit

- **New Features**
  - Inpatient service summaries now list one row per billing instance, including billed date, BHT, and patient details.
  - Staff names are included when available.
  - The inward itemized service summary table is updated to key rows by billing item and to reorder the initial columns to highlight billing and patient information.

- **Bug Fixes**
  - Fixed missing rows caused by nullable patient relationships in JPQL.
  - Prevented fee doubling by relying on stored negative/refund values correctly.

- **Documentation**
  - Added guidance on JPQL nullable-join pitfalls and proper fee sign handling for cancellations/refunds.

### DIFF
--- a/developer_docs/testing/playwright-e2e-workflow.md
+++ b/developer_docs/testing/playwright-e2e-workflow.md
@@ -572,6 +572,31 @@ privilege check for that department under the PO branch's code. Fix: delete (or 
 the rows for privileges that don't exist on the currently deployed branch, re-grant only
 what the current branch's `Privileges.java` actually declares, then re-login.
 
+## 25. A JPQL path expression through a nullable relationship silently INNER-JOINs and drops rows
+
+Writing `b.patientEncounter.bhtNo` or `b.patient.person.name` directly in a `SELECT`/`WHERE`
+generates an **implicit INNER JOIN** on that relationship. If the relationship is null for
+some rows (e.g. `patientEncounter`/`patient` are null on OPD bills), **every one of those
+rows silently disappears** from the result — no error, no log entry. The report just looks
+"wrong" (too few rows), and it's easy to blame filters/dates first.
+
+Tell: a combined OPD+Inward report showed only the 2 Inward rows (which have a
+`patientEncounter`) and dropped all 20 OPD rows for the same item/date range. DB count
+didn't match the report count. Fix: use explicit `left join b.patientEncounter pe` /
+`left join b.patient pat left join pat.person per` and reference the aliases (`pe.bhtNo`,
+`per.name`) in the projection. Verified on issue #21920 (`fetchItemizedServiceInstanceDTOs`
+in `BillService`). Always cross-check report row count against a direct
+`SELECT ... FROM billitem bi JOIN bill b ...` when a report projects fields from an
+optional relationship.
+
+Related sign gotcha found the same pass: cancellation/refund `BillItem` fee columns
+(`netValue`, `hospitalFee`, …) are **already stored negative** in the DB. A
+`case when billClassType in (cancel, refund) then -bi.netValue else bi.netValue end`
+double-negates them to positive — this is the "fee doubling after cancellation" symptom
+(issue #21918). Use the stored values as-is; only synthesize a sign for the row **count**
+(which has no stored value). Verify by summing `bi.netValue` directly in SQL and matching
+the report's Grand Total.
+
 ## Quick checklist
 
 - [ ] Confirmed environment + URL with the developer; credentials kept out of the repo.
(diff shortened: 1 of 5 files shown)


buddhika75:
@coderabbitai review

buddhika75:
Both CodeRabbit findings addressed in `7c292787bf`:

1. **Explicit LEFT JOINs for `bi.item` / `item.category`** — switched to aliases `itm`/`cat` via `left join bi.item itm left join itm.category cat`, so a BillItem with a null item (or an item with a null category) is no longer dropped by an implicit inner join before `coalesce('No Item'/'No Category')` applies. Same pattern applied to the staff-name lookup query.
2. **`bi.retired` guard** — added `and (bi.retired is null or bi.retired=false)` to both the main and staff queries, matching every other item-level query in this file, so retired BillItems on active bills are excluded from the rows and the Grand Total.

Rebuilt, redeployed and re-verified end-to-end: the FBC billed/cancelled window still shows the two entries as separate +456.00 / −456.00 rows and the Grand Total (2,000.00) matches `SUM(bi.netValue)` — no behavior change on existing data.

Issue #21823, opened 2026-07-03
Turn Around Time (Hourly) Report Displays Error Page After Clicking Process Button


chathushkiupulini:
**Describe the bug**
After clicking the Process button, the system renders an error page instead of generating the report.

**To Reproduce**
File path: reports/lab/turn_around_time_hourly.xhtml

1. Navigate to the Turn Around Time (Hourly) Report.
2. Fill in all the required fields with valid data.
43 Click the Process button.

**Expected behavior**
The report should be processed successfully and the Turn Around Time (Hourly) Report should be displayed without any errors.

**Screenshots**

<img width="1903" height="972" alt="Image" src="https://github.com/user-attachments/assets/1ad0a0ea-c9b5-4d4d-953e-4f655a8c744c" />

PR #22296, opened 2026-07-20
#21823 Turn Around Time (Hourly) Report Displays Error Page After Clicking Process Button  

RashmikaHarshamal:


src/main/java/com/divudi/bean/lab/TurnAroundTimeHourlyReport.java
```diff
@@ -0,0 +1,820 @@
...
+import com.itextpdf.text.pdf.PdfPTable;
+import com.itextpdf.text.pdf.PdfWriter;
+
+/**
+ * Turn Around Time (Hourly) Report.
+ *
+ * Groups certified/pending investigations by Investigation + Department and
+ * buckets the elapsed time (order -&gt; result completion) into hourly bands.
+ */
+@Named(value = "tatHourlyReport")
```
buddhika75, line 53: do not overwrite defaults unless essnetial. do we have the same name previously? If not, use the @name without a parameter.

buddhika75 (approved):

Issue #22294, opened 2026-07-20
Baby Admission: Independent discharge (clinical/room/nursing) from parent admission; interim bill scoped to parent only but includes child charges


buddhika75:
## Summary

Baby admissions and their parent (mother's) admission need independent discharge workflows, while interim billing must stay scoped to the parent encounter only — but include all charges accrued under any child (baby) encounters.

## Current Behavior (as of `development`)

- A baby admission is created via `AdmissionController.navigateToAddBabyAdmission()` (`src/main/java/com/divudi/bean/inward/AdmissionController.java:905-921`), which links the new `Admission` to the mother's admission through the generic self-referencing `PatientEncounter.parentEncounter` / `childEncounters` fields (`src/main/java/com/divudi/core/entity/PatientEncounter.java:161,171,524-530`).
- Discharge (clinical / room / nursing) is implemented generically on `PatientEncounter` and operates on whatever encounter is passed to it:
  - Clinical discharge: `InpatientClinicalDataController.confirmClinicalDischarge()` (`InpatientClinicalDataController.java:2937-2971`), page `inward_clinical_discharge.xhtml`
  - Room discharge: `RoomChangeController.java:327-328,366-367`, legacy flow also in `DischargeController.discharge()` (`DischargeController.java:130-144`)
  - Nursing/physical discharge: `NursingDischargeController.confirmNursingDischarge()` / `confirmPhysicalDischarge()` (`NursingDischargeController.java:181,258`), page `inward_nursing_discharge.xhtml`
  - There is currently **no dedicated UI entry point from the Inpatient Dashboard to run these discharge flows against a baby's own admission** — they are reachable for the "current" admission context only, and it's unclear/unverified whether a baby admission can be selected as that context independently of the parent.
- Interim bill generation (`BhtSummeryController.createIntrimBillTable()` / `createTables()`, `BhtSummeryController.java:3133-3175`, page `inward_bill_intrim.xhtml`) already calls `InwardBeanController.fetchChildPatientEncounter()` (JPQL `pe.parentEncounter = :pe and pe.retired=false`) and folds all child-encounter charges (items, pharmacy issues, department bills, outside charges, professional fees) into the bill. Since a baby's `parentEncounter` is the mother's admission, baby charges already flow into the mother's interim bill through this generic mechanism — **but there is no explicit guard preventing an interim bill from being generated directly against a baby (child) encounter itself**, which should not be allowed.
[...]

Issue #20198, opened 2026-04-24
Inward: Baby admission improvements, allergy display on profile, and cancellation UX fixes


buddhika75:
## Summary

A set of related inward/inpatient improvements discovered during development:

---

### 1. Baby / Child Admission

- **Auto-copy guardian details** from the parent (mother's) admission to the baby admission form so staff do not have to re-enter them.
- **Fix back-navigation** from the child admission page: replaced `f:setPropertyActionListener` hack with a proper `navigateCancelBabyAdmission()` method that restores the parent admission before navigating to the Inpatient Dashboard.
- **Skip payment-method validation** for child admissions — child BHTs do not require a separate payment method selection (payment is handled under the parent BHT).

### 2. Patient Allergies on Admission Profile

- **Display allergies panel** on `admission_profile.xhtml` so clinicians can see known allergies at a glance without navigating away.
- **Fix `addPatientAllergy()`** in `BhtEditController`: was adding to the list without persisting; now sets patient reference, sets `ClinicalFindingValueType.PatientAllergy`, and calls `clinicalFindingValueFacade.create()`.
- **Fix `removePatientAllergy()`**: was checking the wrong variable (`currentPatientAllergy` instead of the passed `pa` parameter).
- **Fix allergy JPQL query**: filter by `ClinicalFindingValueType.PatientAllergy` so non-allergy clinical findings are excluded.
- **Smart lazy loading**: track `patientAllergiesLoadedForPatientId` so the list is only reloaded when the patient changes.

### 3. Admission Cancellation UX

- **Clearer error messages** when cancellation is blocked (active bills present, missing comment).
- **Confirmation dialog** (`onclick="return confirm(...)"`) on the Cancel Admission button with the patient name and BHT number so accidental cancellations are prevented.
- **Audit map null-guard**: guard against `NullPointerException` when `originalAdmission` is null before calling `putAll()`; also initialise it in `navigateToEditAdmissionPayment()`.

---

## Files Changed

- `src/main/java/com/divudi/bean/inward/AdmissionController.java`
- `src/main/java/com/divudi/bean/inward/BhtEditController.java`
- `src/main/webapp/inward/admission_profile.xhtml`
- `src/main/webapp/inward/inward_admission_child.xhtml`
- `src/main/webapp/inward/inward_cancel_admission.xhtml`

PR #22300, opened 2026-07-20
Inward: guard interim/estimated bills and enable independent baby admission discharge (#22294)

buddhika75:
## Summary

- Investigation confirmed clinical/room/nursing discharge already work independently for babies today (existing #20198 navigation) — no code change needed there.
- Added a parent-only guard blocking interim/estimated bill generation directly against a baby (child) encounter, at all 4 entry points found (search pages, Room Details button, and a direct-navigation bypass via a lazy getter found during testing).
- Fixed a resulting edge case: since a baby's own administrative `discharged` flag can now never be set directly, `confirmPhysicalDischarge()` now also accepts the parent's `discharged` flag.
- Added a "Baby Admission" badge to the Inpatient Dashboard header so staff can't mistake a baby's page for the mother's.

## Changes

- `BhtSummeryController.java` — parent-only guard in `createIntrimBillTable()`, `createTablesWithEstimatedProfessionalFees()`, `navigateToIntrimBillFromPatientProfile()`, plus a new `preRenderView` guard (`redirectIfEncounterIsBabyAdmission()`) mirroring the existing `redirectIfShiftNotStarted()` precedent, wired into both bill pages.
- `NursingDischargeController.java` — `confirmPhysicalDischarge()` now accepts `currentEncounter.discharged || parentEncounter.discharged`.
- `admission_profile.xhtml` — "Baby Admission — Mother: {name} (BHT {no})" badge, shown only when `parentEncounter != null`.

## Test plan

Verified end-to-end via Playwright against a local build, with a dummy mother/baby admission pair (BHT/56748, BHT/56749) and direct DB inspection — see [issue #22294 comment](https://github.com/hmislk/hmis/issues/22294#issuecomment-5026716919) for screenshots:

- [x] Clinical discharge run independently on the baby — DB confirms baby discharged=1, mother=0
- [x] Direct navigation to the interim bill page while a baby encounter is loaded (browser back/forward bypass) redirects to the dashboard instead of computing baby charges
- [x] Same for the estimated bill page
- [x] Mother's own interim bill still correctly aggregates the baby's room charges (existing aggregation, unchanged)
- [x] Physical discharge on the baby blocked while mother's `discharged=false`
- [x] Physical discharge on the baby succeeds once mother's `discharged=true` — baby's own `discharged` column confirmed to stay `false` throughout (inheritance at the precondition check, not a copy)
- [x] "Baby Admission" badge renders on the baby's dashboard, absent on the mother's

Design spec and implementation plan committed to `docs/superpowers/` for reference.

### DIFF
--- a/src/main/java/com/divudi/bean/inward/NursingDischargeController.java
+++ b/src/main/java/com/divudi/bean/inward/NursingDischargeController.java
@@ -268,7 +268,10 @@ public class NursingDischargeController implements Serializable {
             JsfUtil.addErrorMessage("Cannot confirm physical discharge: nursing discharge has not been completed.");
             return;
         }
-        if (!Boolean.TRUE.equals(currentEncounter.getDischarged())) {
+        boolean administrativelyDischarged = Boolean.TRUE.equals(currentEncounter.getDischarged())
+                || (currentEncounter.getParentEncounter() != null
+                    && Boolean.TRUE.equals(currentEncounter.getParentEncounter().getDischarged()));
+        if (!administrativelyDischarged) {
             JsfUtil.addErrorMessage("Cannot confirm physical discharge: administrative discharge (final bill) has not been completed.");
             return;
         }
--- a/docs/superpowers/plans/2026-07-20-baby-admission-independent-discharge.md
+++ b/docs/superpowers/plans/2026-07-20-baby-admission-independent-discharge.md
@@ -310,7 +310,102 @@ EOF
 
 ---
 
-### Task 4: Full build, redeploy, and Playwright + DB verification
+### Task 4: Guard the interim/estimated bill pages against direct navigation with a baby encounter loaded
+
+**Files:**
+- Modify: `src/main/java/com/divudi/bean/inward/BhtSummeryController.java` (new method, placed immediately before `createIntrimBillTable()` at line 3133)
+- Modify: `src/main/webapp/inward/inward_bill_intrim.xhtml` (add `f:event` inside `f:metadata` near the top of the page)
+- Modify: `src/main/webapp/inward/inward_bill_intrim_estimate.xhtml` (same)
+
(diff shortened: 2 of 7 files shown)


src/main/java/com/divudi/bean/inward/NursingDischargeController.java
```diff
@@ -268,7 +268,10 @@ public void confirmPhysicalDischarge() {
             JsfUtil.addErrorMessage("Cannot confirm physical discharge: nursing discharge has not been completed.");
             return;
         }
-        if (!Boolean.TRUE.equals(currentEncounter.getDischarged())) {
+        boolean administrativelyDischarged = Boolean.TRUE.equals(currentEncounter.getDischarged())
+                || (currentEncounter.getParentEncounter() != null
+                    && Boolean.TRUE.equals(currentEncounter.getParentEncounter().getDischarged()));
```
buddhika75, line 273: **Bug: `BhtSummeryController.discharge()` (final bill / administrative discharge) has no parent-only guard, unlike its three sibling methods.**

This change makes `confirmPhysicalDischarge()` treat a baby as administratively discharged once **its parent's** `discharged` flag is true — which quietly assumes the invariant "a baby's own `discharged` flag can never be set directly" (per the PR description). But nothing in this PR actually enforces that invariant at the source: `BhtSummeryController.discharge()` (which sets `patientEncounter.setDateOfDischarge(...)` and ultimately `discharged=true` via `DischargeController.discharge()`) still operates on whatever `patientEncounter` happens to be, with no `patientEncounter.getParentEncounter() != null` check — unlike `createIntrimBillTable()`, `createTablesWithEstimatedProfessionalFees()`, and `navigateToIntrimBillFromPatientProfile()`, which all got this guard in this same PR.

`discharge()` is invoked from the "Discharge" button on the very two pages (`inward_bill_intrim.xhtml:386`, `inward_bill_intrim_estimate.xhtml:213`) that this PR also added the `redirectIfEncounterIsBabyAdmission()` preRenderView guard to — but that guard only fires on the initial GET and explicitly skips postbacks (`fc.isPostback()`). Since `BhtSummeryController` is `@SessionScoped`, `patientEncounter` is shared across all browser tabs for the same logged-in user. Concrete failure scenario:

1. Tab A: staff generates the mother's interim bill (`patientEncounter` = mother), sees the results panel with the "Discharge" button rendered.
2. Tab B (same session/browser): staff navigates to the baby's own Inpatient Dashboard — this unconditionally mirrors `admissionController.current` (the baby) into `bhtSummeryController.patientEncounter` (per this PR's own investigation notes), silently overwriting the session-scoped field.
3. Back in Tab A (still showing the mother's stale page), staff clicks "Discharge". This is a postback, so the preRenderView guard is skipped, and `discharge()` reads the *current* session-scoped `patientEncounter` — now the baby — and discharges/finalizes billing on the baby directly, with no guard to stop it.

This defeats the "administrative discharge is parent-only" design this PR is built around, and directly undermines the assumption Task 2's `confirmPhysicalDischarge()` fix relies on. Recommend adding the same `patientEncounter.getParentEncounter() != null` guard directly inside `discharge()` (around `BhtSummeryController.java:2429`), consistent with the defense-in-depth already applied to the three sibling methods in this PR.

buddhika75: Fixed in 21fccbb00d: added the same `patientEncounter.getParentEncounter() != null` guard directly inside `discharge()` (BhtSummeryController.java:2429), matching the pattern already used in the three sibling methods. Confirmed the cross-tab scenario you described — the preRenderView guard only fires on GET and skips postbacks, so `discharge()` itself needed its own guard rather than relying on the page-load check. Good catch.

PR #22419, opened 2026-07-25
Add fullAccess attribute to restrict admin toolset on unauthenticated migration page

buddhika75:
## Summary
Introduces a `fullAccess` boolean attribute to the `midding_data_fields` composite component to conditionally expose either a simplified single-button DDL interface or the full four-tab admin toolset. This allows the unauthenticated migration-pending page (`mf.xhtml`) to safely display database update functionality without exposing sensitive admin operations.

## Key Changes

- **Composite Component (`midding_data_fields.xhtml`)**:
  - Added `fullAccess` attribute (default `true`) to control UI visibility
  - Full "Add All Tables" panel now renders only when `fullAccess="true"`
  - New simplified "Add All Tables" panel renders only when `fullAccess="false"` with a single button: "Load Latest DDL from Wiki and Update Both Databases"
  - "Check Missing Fields and Add Fields" tab now conditionally rendered based on `fullAccess`
  - "Run Custom SQL Commands" tab now conditionally rendered based on `fullAccess`

- **Controller (`DataAdministrationController.java`)**:
  - Added `loadDdlFromWikiAndUpdateBothDatabases()` method to support the simplified single-button workflow
  - This method automatically sets both database flags and triggers the DDL load/apply in one step

- **Migration Page (`mf.xhtml`)**:
  - Updated to pass `fullAccess="false"` to the composite component
  - Added clarifying comments explaining the reduced toolset for unauthenticated access
  - Changed non-migration-pending panel to display an informational message instead of exposing the component

- **Admin Page (`missing_database_fields.xhtml`)**:
  - Added authorization check to ensure only users with 'Admin' privilege can access the full toolset
  - Passes `fullAccess="true"` to expose all four tabs and admin functionality

## Implementation Details

The simplified interface (when `fullAccess="false"`) provides a single, self-contained button that downloads the latest DDL from the wiki and applies it to both main and audit databases without requiring manual database selection or SQL snippet pasting. This maintains security by preventing unauthenticated users from accessing the "Check Missing Fields" and "Run Custom SQL Commands" tabs, which remain available only through the authenticated Admin Functions path.

https://claude.ai/code/session_019uMkwty178PBBLLboTKr5B

## Summary by CodeRabbit

* **New Features**
  * Added a single-action option to load the latest database definition and update both databases.
  * Added simplified database migration controls with separate execution feedback for each database.
  * Added clear messaging when no database migration is pending.

* **Access Control**
  * Restricted database administration tools and custom SQL execution to authorized administrators.
  * Displayed an authorization message for users without administrator access.

### DIFF
--- a/src/main/java/com/divudi/bean/common/DataAdministrationController.java
+++ b/src/main/java/com/divudi/bean/common/DataAdministrationController.java
@@ -2401,6 +2401,21 @@ public class DataAdministrationController implements Serializable {
     }
 
     /**
+     * Single-click path for the reduced, unauthenticated "Add Missing Fields"
+     * page (mf.xhtml during pending migration): downloads the latest DDL from
+     * the wiki and applies it to both the main and audit databases, with no
+     * manual database-selection checkboxes or paste-a-snippet option — those
+     * remain available only in the full admin tool.
+     */
+    public void loadDdlFromWikiAndUpdateBothDatabases() {
+        allCreateStetements = "";
+        runOnMainDatabase = true;
+        runOnAuditDatabase = true;
+        loadDdlFromWiki();
+        createTablesAndFieldsForAllCreateStatements();
+    }
+
+    /**
      * The DDL to operate on: manually pasted content (Tab 1 textarea) wins;
      * otherwise falls back to the script fetched server-side from the wiki.
      */
--- a/src/main/java/com/divudi/bean/common/DataAdministrationController.java
+++ b/src/main/java/com/divudi/bean/common/DataAdministrationController.java
@@ -2412,6 +2412,12 @@ public class DataAdministrationController implements Serializable {
         runOnMainDatabase = true;
         runOnAuditDatabase = true;
         loadDdlFromWiki();
+        if (wikiFetchedDdl == null || wikiFetchedDdl.trim().isEmpty()) {
+            // loadDdlFromWiki() already set executionFeedback with the specific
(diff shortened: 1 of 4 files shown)


buddhika75:
Pushed a follow-up commit addressing the CodeRabbit review:

- **Fixed (security, major)**: the reduced/unauthenticated feedback panels used `escape="false"` with a raw `<textarea>`, which would let `executionFeedback`/exception text break out via a literal `</textarea>` and inject markup on a page reachable without login. Switched to `p:inputTextarea`, matching the pattern already used elsewhere in this file.
- **Fixed**: `loadDdlFromWikiAndUpdateBothDatabases()` now aborts before running the update step if the wiki DDL download failed, instead of overwriting the specific download error with a generic "DDL content is empty" message.
- **Fixed (minor)**: `p:outputLabel` → `h:outputText` for the "not authorized" message; explicit `layout="block"` on the no-migration-pending panel in `mf.xhtml`.
- **Skipped**: "only mark migration schema version current after both databases succeed" — that logic lives in the pre-existing `createTablesAndFieldsForAllCreateStatements()`/`markSchemaAsCurrentSilently()`, unchanged by this PR and shared with the existing admin "Update Database" button. Reworking per-database success tracking there is a larger, unrelated correctness change I'd rather not bundle into a security fix — happy to file a separate issue for it if wanted.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #22548, opened 2026-07-30
feat(pharmacy): add substitute option to Direct Issue to BHT

### LINKED ISSUE #22482: Substitute Option Missing for Direct Issue to BHT (Pharmacy and Inward)
File Path: inward/pharmacy_bill_issue_bht.xhtml
             
Description:
The Substitute option is not available in the Direct Issue to BHT module for both Pharmacy and Inward. Users are unable to substitute an item when the originally requested item is unavailable.

Steps to Reproduce:

1. Navigate to Direct Issue to BHT (Pharmacy or Inward).
2. Select a BHT issue request.
3. Attempt to substitute an item.
4. Observe the available actions.

Actual Result:

The Substitute option is missing.

Expected Result:

A Substitute option should be available in Direct Issue to BHT for both Pharmacy and Inward.

buddhika75:
## Summary
- Adds a "Select a Substitute" action to `inward/pharmacy_bill_issue_bht.xhtml` (Direct Issue to BHTs — reached from both the Pharmacy and Inward menus, a single screen not two separate implementations), matching the alternatives feature already available on Retail Sale. Closes #22482.
- `InpatientDirectIssueNativeSqlController` gains `prepareSubstitute()` / `replaceSelectedSubstitute()`, reusing the shared native-SQL `PharmacySubstituteService.findSubstituteStocks()` lookup (FEFO-ordered, in-stock, non-expired sibling AMPs under the same VMP) already powering Retail Sale's substitute feature — no new business logic for stock resolution, just wiring this DTO-based controller into the existing shared service.
- Price-matrix margin/discount computation was extracted from `addBillItem()` into a small `computeMarginAndDiscount()` helper so the substitute swap recalculates pricing exactly the way a normal add does (this controller needs both margin *and* discount, unlike Retail Sale's simpler discount-only substitute logic).
- Package-priced lines (`bid.fromPackage`) are excluded from substitution — button is replaced with a lock icon/tooltip, and the Java methods reject the action defensively too — since the package quota was reserved for the original item specifically and swapping items would silently orphan that reservation.
- No config toggle (unlike Retail Sale's) — the button always renders; the page is already privilege-gated. No entity/DTO/schema changes; `BillItemData` already carried every field the swap needed.

## Test plan
Verified end-to-end with Playwright against the local dev DB (Main Pharmacy department, BHT/56750):
- [x] Added a bill line for an item with in-stock sibling brands under the same VMP — substitute button appeared on the row.
- [x] Opened the Alternatives dialog — FEFO-ordered sibling brands listed with stock/rate/expiry, paginated.
- [x] Replaced the line with a different sibling brand — item name, rate, gross/service-charge/discount/net values, and expiry all updated correctly in the UI, bill totals recalculated.
- [x] Settled the bill and confirmed directly in the database: the `PHARMACEUTICALBILLITEM` row for the line points at the substituted item's stock/batch (not the original), that batch's stock was decremented by the issued quantity, and the original item's stock was untouched.
- [x] Reviewed the package-priced exclusion via code (both the XHTML `rendered` conditions and the Java `isFromPackage()` guards) — did not stand up a full inpatient-package test encounter locally since that would be a large detour for a two-line boolean guard.

Screenshots and full verification notes: https://github.com/hmislk/hmis/issues/22482#issuecomment-5127461790

### DIFF
--- a/src/main/java/com/divudi/bean/pharmacy/InpatientDirectIssueNativeSqlController.java
+++ b/src/main/java/com/divudi/bean/pharmacy/InpatientDirectIssueNativeSqlController.java
@@ -27,6 +27,7 @@ import com.divudi.core.util.JsfUtil;
 import com.divudi.ejb.BillNumberGenerator;
 import com.divudi.ejb.PharmacyService;
 import com.divudi.service.pharmacy.InpatientDirectIssueNativeSqlService;
+import com.divudi.service.pharmacy.PharmacySubstituteService;
 import com.divudi.service.pharmacy.PriceMatrixNativeSqlService;
 import com.divudi.core.util.CommonFunctions;
 
@@ -88,6 +89,8 @@ public class InpatientDirectIssueNativeSqlController implements Serializable {
     @EJB
     private PriceMatrixNativeSqlService priceMatrixNativeSqlService;
     @EJB
+    private PharmacySubstituteService pharmacySubstituteService;
+    @EJB
     private com.divudi.core.facade.BillItemFacade billItemFacade;
     @EJB
     private com.divudi.core.facade.InpatientPackageItemFacade inpatientPackageItemFacade;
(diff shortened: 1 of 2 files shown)


buddhika75:
@coderabbitai review

Issue #21266, opened 2026-06-05
fix(pharmacy): duplicate BillItems on transfer receive due to JPA cascade double-insert


buddhika75:
## Summary

When a pharmacy transfer receive is settled via `TransferReceiveController.settle()`, the receiving department gets **double stock-in** for every item. In the stock ledger, the same bill number appears twice with identical quantities and timestamps.

**Reported from**: Ruhunu Hospital — Transfer Request `TREQ/RHD/LBK/26/00074`, Receive Bill `TR/RHD/LBK/26/00075`
**Item affected**: Caresens N Strips — received 30 units twice (60 total) when only 30 should have been received.

---

## Root Cause

`Bill.billItems` is mapped with `cascade = CascadeType.ALL`:

```java
@OneToMany(mappedBy = "bill", cascade = CascadeType.ALL, fetch = FetchType.LAZY, orphanRemoval = true)
private List<BillItem> billItems;
```

In `settle()` (`TransferReceiveController.java`):

1. `generateBillComponent()` populates `receivedBill.billItems` in memory (all `id == null`)
2. `saveBill()` calls `getBillFacade().create(receivedBill)` — JPA cascade **persists all BillItems automatically** (they now have IDs)
3. The explicit loop then checks `if (i.getId() == null)` — they are NOT null, so it goes to the `edit()` branch and calls `addToStock()` once (correct)
4. `getBillFacade().edit(getReceivedBill())` at **line 564** merges the `@SessionScoped` bean's `receivedBill` back — EclipseLink's merge + `CascadeType.ALL` **re-inserts BillItems as new rows** because after the `create()` transaction committed, the session-scoped bean holds detached references no longer tracked by the persistence context
5. `getBillFacade().edit(getReceivedBill())` is called **again at line 569** (redundant second call)

Result: each BillItem appears twice in the DB under the same bill, making the stock ledger show a double stock-in.

---

### 1. `TransferReceiveController.saveBill()` — prevent cascade on initial create

Save the bill header with an empty items list so JPA cascade does not fire prematurely. The explicit loop in `settle()` already handles individual item creation/editing:

```java
public void saveBill() {
    ...
    if (getReceivedBill().getId() == null) {
        List<BillItem> items = getReceivedBill().getBillItems();
        getReceivedBill().setBillItems(new ArrayList<>());
        getBillFacade().create(getReceivedBill());
        getReceivedBill().setBillItems(items);
    } else {
        getBillFacade().edit(getReceivedBill());
[...]

PR #22939, opened 2026-08-14
fix(pharmacy): abort transfer issue settlement on mid-loop stock shortfall

### LINKED ISSUE #22938: Pharmacy Transfer Issue: insufficient-stock check silently skips stock movement, leaves phantom BillItem
## Summary

`TransferIssueForRequestsController.settle()` (Main Pharmacy -> department Transfer Issue flow) silently skips the actual stock movement for a line item when live stock is insufficient, but still persists the BillItem/PharmaceuticalBillItem as if the full requested quantity was issued. No error is shown to the user and the settlement is not aborted.

## Production incident (coop, 2026-08-13)

Bill `TI/COOP/26/000488` (ID 14048774, Main Pharmacy -> OPD Pharmacy, 30 line items) was issued and a paper Transfer Issue Note was printed and signed. One line — Evion 400mg capsule, itembatch 13860470, requested qty 141 — has:
- `BillItem`/`PharmaceuticalBillItem` persisted with `QTY=-141` (appears on the bill and printed note with full value)
- `PharmaceuticalBillItem.STAFFSTOCK_ID = NULL` (no porter-holding record created)
- Zero `StockHistory` rows (no deduction from Main Pharmacy, no credit to OPD Pharmacy)

Main Pharmacy's actual stock for that batch was exactly 140 units one minute before the transaction — one short of the 141 requested. Neither department's stock actually moved for this item, but the bill and printed note both show it as transferred.

## Root cause

`src/main/java/com/divudi/bean/pharmacy/TransferIssueForRequestsController.java`, `settle()`:

- A pre-loop stock check (~line 431-435) validates against a cached `Stock.getStock()` value that can be stale.
- Inside the per-item persistence loop, after the `BillItem`/`PharmaceuticalBillItem` are already `create()`d, a second live check `userStockController.isStockAvailable(...)` (~line 500) correctly detects the real shortfall and returns `false`. On that `false`:
  ```java
  if (!userStockController.isStockAvailable(tmpPh.getStock(), tmpPh.getQty(), getSessionController().getLoggedUser())) {
      billItemsInIssue.setTmpQty(0);
      getBillItemFacade().edit(billItemsInIssue);
      getIssuedBill().getBillItems().add(billItemsInIssue);
      continue;
  }
  ```
  This sets a cosmetic `tmpQty=0` and `continue`s to the next item — no error message, no abort, and the already-persisted BillItem keeps its original (unfulfillable) `qty`.
- Similarly, when `pharmacyBean.deductFromStock(...)` returns `false` (its own internal `stock.getStock() < qty` guard), the `else` branch (~line 547-548) just `edit()`s the BillItem with no error and no abort.

Neither failure path raises an exception — both are boolean-return checks that fall through to a no-op, so there's nothing for a `try/catch` to catch, and no transactional rollback undoes the already-created `BillItem`/`PharmaceuticalBillItem`.

## Related precedent

This is the same bug class fixed on the **receive** side of the same porter-mediated transfer feature: commit `7d4810a8bf` / `e72a1c37ed` ("stop phantom transfer receives bound to issuing dept stock", #21266 RC2) fixed `TransferReceiveController.java` for an equivalent Ruhunu production incident (12 bills, 75 lines) — but that fix never covered `TransferIssueForRequestsController.java`, which has the same flawed pattern on the **issue** side.

Also present in `development` branch (confirmed via diff) — not coop-specific.

## Fix approach (mirroring `790cc3c4b6` / `3c2660e47d`, the OPD batch-bill silent-persistence-failure hotfix)

When the stock-availability/deduction check fails for an item inside `settle()`:
- Raise a clear `JsfUtil.addErrorMessage(...)` naming the item and shortfall
- Abort the whole settlement (`return`) instead of `continue`/silent no-op
- Retire any `BillItem`/`PharmaceuticalBillItem`/`Bill` rows already persisted earlier in that same `settle()` call, so a failed settlement doesn't leave an orphaned partial bill (same pattern as `retirePartialBillOnSettlementFailure()` in `OpdBillController.java`)

## Data correction needed (coop production)

Bill 14048774's Evion 400mg line needs its quantity corrected from -141 to -140 (true available stock) and the corresponding stock movement (StaffStock + StockHistory, Main Pharmacy deduction + OPD Pharmacy credit) performed retroactively, since it never happened.

buddhika75:
## Summary
- `TransferIssueForRequestsController.settle()` silently kept going when `deductFromStock()` failed mid-loop for a line item (live stock dropped below the requested qty between the pre-loop check and that item's turn) — the `else` branch just `edit()`d the already-persisted BillItem with no error and no stock movement, leaving a phantom line on the bill.
- Adds an abort path: on failure, retire the Bill and any BillItems already persisted in the same `settle()` call (including the failing item itself), show a clear error, and stop the settlement instead of continuing.

## Production incident
Bill `TI/COOP/26/000488` (coop, 13 Aug 2026): Evion 400mg capsule requested at qty 141, but Main Pharmacy only had 140 in stock. The line was saved with qty -141 and shown on the printed Transfer Issue Note, but `STAFFSTOCK_ID` stayed NULL and zero `StockHistory` rows were written — neither side's stock actually moved. Production data already corrected separately via `BillDataCorrectionApi` + `PharmacyAdjustmentApi` (qty -141 → -140, stock adjustment bills `MP//26/046293`/`OP//26/046294`).

## Related precedent
Same bug class as:
- `790cc3c4b6`/`3c2660e47d` — OPD batch-bill silent BillItem persistence failure guard
- `7d4810a8bf`/`e72a1c37ed` — Ruhunu transfer-**receive** phantom-stock fix (#21266 RC2)

The receive-side controller (`TransferReceiveController`) got this guard for the Ruhunu incident; the issue-side controller (`TransferIssueForRequestsController`) never did, until now.

Closes #22938

## Test plan
- [x] `mvn compile` passes clean
- [ ] Reviewer: confirm the normal single/multi-item settle happy path is unaffected (no behavior change when stock is sufficient)
- [ ] Reviewer: confirm the abort path retires Bill + prior BillItems and shows the error message when reproduced with a forced insufficient-stock scenario

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stock transfer failure handling when insufficient stock is detected.
  * Prevented partially completed transfer issues from remaining after an unsuccessful settlement.
  * Clearer failure reporting now appears when stock movement cannot be completed.

### DIFF
--- a/src/main/java/com/divudi/bean/pharmacy/TransferIssueForRequestsController.java
+++ b/src/main/java/com/divudi/bean/pharmacy/TransferIssueForRequestsController.java
@@ -794,6 +794,37 @@ public class TransferIssueForRequestsController implements Serializable {
         JsfUtil.addSuccessMessage("Issue draft cancelled.");
     }
 
+    /**
+     * Retires the in-progress issue Bill and any BillItems already persisted
+     * earlier in the current settle() call (including the item that just
+     * failed, which is not yet added to bill.getBillItems() at the point of
+     * failure), so a mid-loop stock-movement failure does not leave an
+     * orphaned partial bill (mirrors
+     * OpdBillController.retirePartialBillOnSettlementFailure(), #22399/#22938).
+     */
+    private void retirePartialIssueOnSettlementFailure(BillItem failedItem) {
+        Bill bill = getIssuedBill();
+        if (bill == null || bill.getId() == null) {
+            return;
+        }
+        List<BillItem> toRetire = new ArrayList<>(bill.getBillItems());
+        if (failedItem != null && failedItem.getId() != null) {
+            toRetire.add(failedItem);
+        }
+        for (BillItem bi : toRetire) {
+            if (bi != null && bi.getId() != null && !bi.isRetired()) {
+                bi.setRetired(true);
+                bi.setRetireComments("Transfer issue settlement aborted: insufficient stock mid-settlement");
+                getBillItemFacade().edit(bi);
+            }
+        }
+        if (!bill.isRetired()) {
+            bill.setRetired(true);
+            bill.setRetireComments("Transfer issue settlement aborted: insufficient stock mid-settlement");
+            getBillFacade().edit(bill);
+        }
+    }
+
     public synchronized void settle() {
         if (!isAuthorized("SETTLE_ISSUE", "PharmacyIssueForRequestFinalize")) {
             return;
@@ -952,7 +983,20 @@ public class TransferIssueForRequestsController implements Serializable {
                 getBillItemFacade().edit(billItemsInIssue);
                 getBillItemFacade().edit(originalOrderItem);
             } else {
-                getBillItemFacade().edit(billItemsInIssue);
+                // Stock dropped below the requested qty between the pre-loop live check
+                // and this item's turn in the loop (e.g. another line in the same settle()
+                // call consumed the same batch). deductFromStock() did NOT move any stock
+                // or write StockHistory here - continuing would leave this BillItem/
+                // PharmaceuticalBillItem persisted with a qty that was never actually
+                // transferred (coop production, TI/COOP/26/000488, Evion 400mg capsule,
+                // #22938). Abort and retire whatever this settle() already persisted
+                // instead of silently keeping a phantom line.
+                String name = billItemsInIssue.getItem() != null ? billItemsInIssue.getItem().getName() : "one of the items";
+                retirePartialIssueOnSettlementFailure(billItemsInIssue);
+                JsfUtil.addErrorMessage("Insufficient stock to issue " + name + ". "
+                        + "Stock levels may have changed since this page was loaded. "
+                        + "No items were issued - please refresh and try again.");
+                return;
             }
 
             getPharmaceuticalBillItemFacade().edit(billItemsInIssue.getPharmaceuticalBillItem());
--- a/src/main/java/com/divudi/bean/pharmacy/TransferIssueForRequestsController.java
+++ b/src/main/java/com/divudi/bean/pharmacy/TransferIssueForRequestsController.java
@@ -795,34 +795,75 @@ public class TransferIssueForRequestsController implements Serializable {
     }
 
     /**
-     * Retires the in-progress issue Bill and any BillItems already persisted
-     * earlier in the current settle() call (including the item that just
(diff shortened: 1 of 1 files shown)


buddhika75:
Updated the approach after reviewing feedback that surfaced real gaps in the retire-based rollback (stock movements for earlier items in the same `settle()` call weren't reversed, source request's `remainingQty` desynced, `PharmaceuticalBillItem` never retired, no in-page retry path). A correct rollback would need to unwind Bill/BillItem/PharmaceuticalBillItem/Stock/StockHistory/StaffStock/originalOrderItem consistently — too much surface for a hotfix.

Replaced it with prevention instead: `findInsufficientStockError()` aggregates every line's requested qty per Stock row across the whole `settle()` call and validates live DB stock can cover the combined total, before any persistence starts. This is exactly the shape of the actual bug (one line checked in isolation against a batch that couldn't cover it) and also closes the case of two-or-more lines splitting the same batch, which the old per-line check couldn't see.

The narrow residual TOCTOU race is now handled by logging for manual reconciliation (`logResidualStockRace()`) rather than attempting a rollback after persistence has already started — follow-up for proper Notification delivery on that rare case tracked in #22942.

buddhika75:
Follow-up fix pushed (commit bfcce69e9b): the receive side had its own crash bug, separate from the issue-side fix in this PR.

**Root cause (confirmed via coop production Payara log):** \`TransferReceiveController.errorCheck()\` dereferenced \`pharmaceuticalBillItem.getStaffStock().getStock()\` without a null check. When an issue-side line never actually moved stock to the porter (this PR's bug), the receive-side PBI inherits a null \`staffStock\`, and \`errorCheck()\` throws an uncaught NPE - crashing \`settle()\` mid-loop and leaving the request partially processed with no error shown.

This is exactly what happened on TI/COOP/26/000488: the NPE at 2026-08-13T15:34:04 on the Evion line left 16 of 30 items stuck at the porter with no stock movement to OPD Pharmacy, despite the receive bill showing them as received.

Fixed both \`errorCheck()\` and \`onEdit()\` with a null guard, plus a pre-loop check ahead of \`saveBill()\` so a bad line is reported clearly before persistence starts.

Tracked in #22951. (Note: earlier commit message on this branch referenced "#22943" - that was a placeholder before the issue was actually filed; the correct number is #22951.)

PR #21721, opened 2026-06-28
fix(inward): exclude refunded lab items from inpatient lab investigation list [coop-prod]

buddhika75:
## Summary
Hotfix for coop-production — cherry-pick of #21719 (merged to `development`) and #21720 (merged to `coop-stg`).

Refunded lab investigation items were still appearing in the inpatient lab investigation list report at `/inward/reports/inpatient_lab_investigation_item_list.xhtml`.

`fetchLabBillItems()` JPQL was missing `bi.refunded = false`, `bi.retired = false`, and `bi.bill.cancelled = false` — so refunded items were not excluded from the list.

## Changes
Single file: `InwardReportControllerBht.java` — adds three filter conditions to the JPQL in `fetchLabBillItems()`.

Closes #21718

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #21718, opened 2026-06-28
fix(inward): refunded lab items still appear in inpatient lab investigation list report


buddhika75:
## Problem

After a refund is processed at `/inward/inward_bill_service_refund.xhtml`, the refunded investigation items still appear in the inpatient lab investigation list report at `/inward/reports/inpatient_lab_investigation_item_list.xhtml`.

## Root Cause

`InwardReportControllerBht.fetchLabBillItems()` queries only for `bi.bill.retired = false` but does not filter on `bi.refunded` or `bi.retired` at the `BillItem` level.

When `InwardServiceRefundController` processes a refund, it correctly sets `original.setRefunded(true)` on each refunded `BillItem`. However, since the report query ignores this flag, refunded items continue to appear.

This must be fixed at the `BillItem` level (not bill level) because a single bill can have 10 items where only some are refunded.

## Fix

Add `AND bi.refunded = false AND bi.retired = false` to the JPQL in `fetchLabBillItems()` in `InwardReportControllerBht.java`.

PR #21422, opened 2026-06-11
hotfix(pharmacy): coop — PO request PBI fixes + stock-integrity fixes from #21401

buddhika75:
## Summary

Hotfix for **coop production** combining two fix sets:

1. **PO request duplicate-billitem / zero-qty PBI fix** (issue #21417, merged to `development` via #21420) — the bug that blocked GRN for **POR/COOP/MP/26/000965** (Maxgalin 50mg qty 0; incident record #21418)
2. **Pharmacy stock-integrity fixes from the Ruhunu hotfix #21401** (#21266 family — all 6 root causes), cherry-picked onto `coop-prod`

## Part 1 — PO request fix (cherry-pick of #21420, commit `b544eb2d58`)

- `saveRequestWithoutMessage()` / `finalizeRequest()` synchronized — concurrent saves from one session persisted the same in-memory BillItem twice (duplicate rows sharing one BIFD)
- `resyncPharmaceuticalBillItemIfEmpty()` at save/finalize — rebuilds an all-zero PBI from the finance details so a PO can no longer reach approval/GRN with qty 0
- Enter-key guards on the PO request page (qty/price inputs + autocompletes)

## Part 2 — stock-integrity fixes (cherry-picked from #21401)

| Fix | Commit (this branch) |
|---|---|
| Disposal Consumption Report: net consumption | `21b4f86fc6` |
| Block GRN Return approval when stock insufficient | `8ac5f7b333` |
| Pre-deduction stock check in `updateStock()` | `6eb34746b1` |
| Duplicate BillItems on transfer receive (cascade re-insert) + a11y | `30ca20091e` |
| Transfer receive review fixes (settling guard, edit-after-fillData, try/finally restore) | `9c76b53fc6` |
| COGS report: date by stock-move date, remove double-counts | `2d85c5c26d` |
| Phantom transfer receives bound to issuing dept stock | `388775defb` |
| Stale-list Close clobbering settled disposal returns | `8b7993b077` |
| 36 missing Privileges enum constants (General Stores login) | `df77820070` |
| Transfer issue: move stock by user-entered qty | `7428587104` |
| Orphan-removal guard on disposal return merges | `fa4478ac97` |
| Status panels explaining hidden disposal-return buttons | `8a6ad187a1` |
| GRN return approval: exact finalized quantities only | `0083304f0e` |
| Direct purchase return approval: exact finalized quantities only | `8965b065c6` |
| Count stock-take adjustment bills in COGS | `607934adc9` |
[...]

Issue #21417, opened 2026-06-11
PO request double-submit creates duplicate BillItem sharing one BIFD; surviving line ends with zero-qty PharmaceuticalBillItem, blocking GRN


buddhika75:
## Summary

A double-submit race on the Purchase Order Request page (`pharmacy_purhcase_order_request.xhtml` / `PurchaseOrderRequestController`) can create **duplicate `BillItem` rows that share a single `BillItemFinanceDetails`**, with only one of them owning the `PharmaceuticalBillItem`. When a user later removes the visible duplicate line, the row holding the real PBI may be the one retired — leaving the surviving line with a lazily-created, all-zero PBI. Downstream PO approval / GRN reads quantity from the PBI, so the item shows **qty 0 and cannot be GRN'd**.

## Real-world occurrence (coop production)

PO Request **POR/COOP/MP/26/000965** (bill 13915155), item **Maxgalin 50mg**:

| Record | Value |
|---|---|
| billitem 13998848 (retired 2026-05-27) | qty 300, owned the real PBI 13979280 (qty 300, rate 20.15) |
| billitem 13998849 (live) | qty 300, BIFD qty 300, but PBI 14005349 had **qty 0, rate 0, CREATEDAT NULL** |
| Shared BIFD | both billitems pointed to BIFD 13958481 family row 13958492 |

Timeline:
1. **2026-05-21 12:57:59** — both billitems created in the same second, sharing one BIFD, only one PBI created. A shared BIFD proves the same in-memory object graph was persisted twice (double-submit race), not two separate UI adds.
2. **2026-05-27 14:26** — user edited the request and removed the duplicate line; the removed row was the one holding the real PBI (retired together with it, `removeItem()` behaviour).
3. The surviving billitem had no PBI; `BillItem.getPharmaceuticalBillItem()` lazily creates an empty PBI when null, so an all-zero PBI was silently persisted on the next save (bill edited 2026-05-29). `calculateLineValues()` never ran for that line because the user never re-typed its qty, so the zeros stuck.
4. GRN blocked: `PurchaseOrderController` reads `pharmaceuticalBillItem.getQty()` (≈ lines 525–530) → 0.

The production data was corrected directly (see companion production issue). This issue is for the **code fixes**.

## Proposed fixes

1. **Guard the Enter key / double-submit on the PO request page** — same defaultCommand race fixed for direct purchase in #21415. Apply the identical guard pattern to `pharmacy_purhcase_order_request.xhtml`.
[...]

PR #22940, opened 2026-08-14
fix(pharmacy): coop-prod hotfix bundle — transfer issue stock skip, PO request PBI guard, refunded lab items

### LINKED ISSUE #22938: Pharmacy Transfer Issue: insufficient-stock check silently skips stock movement, leaves phantom BillItem

buddhika75:
## Summary
Coop-prod hotfix bundle — combined into one PR to minimize downtime windows.

### 1. Transfer issue stock skip (original scope, #22938)
Coop-prod companion to #22939 (development). This branch's `settle()` has two silent-skip points instead of one (older, diverged code):
- `isStockAvailable()` returning `false` zeroed `tmpQty` and `continue`d — no error shown.
- `deductFromStock()` returning `false` fell into a no-op `else` — no error shown.

Both now abort the whole settlement with a clear error message and retire the Bill + any BillItems already persisted earlier in the same `settle()` call.

**Production incident:** Bill `TI/COOP/26/000488` (13 Aug 2026): Evion 400mg capsule requested at qty 141, Main Pharmacy only had 140 in stock. Line persisted with qty -141, `STAFFSTOCK_ID` NULL, zero `StockHistory` rows — neither side's stock moved, but the bill/printed note showed it transferred. Data already corrected via `BillDataCorrectionApi` + `PharmacyAdjustmentApi` (see #22938 for full detail).

### 2. PO request duplicate BillItem / zero-qty PBI guard (#21417)
A double-submit race on the purchase order request page (Enter-key `defaultCommand` firing btnSave while another save was in flight) persisted the same in-memory BillItem twice, creating duplicate rows that share one BillItemFinanceDetails with only one of them owning the PharmaceuticalBillItem. Removing the visible duplicate could retire the row holding the real PBI; the survivor then got a lazily created all-zero PBI, and PO approval/GRN (which reads `pbi.qty`) showed qty 0 and blocked the GRN (coop `POR/COOP/MP/26/000965`, Maxgalin 50mg).

- `saveRequestWithoutMessage()` / `finalizeRequest()` now `synchronized`, serializing concurrent saves from one session
- `resyncPharmaceuticalBillItemIfEmpty()` rebuilds an all-zero PBI from finance details, self-healing already-corrupted lines
- Page-side Enter-key guards on qty/free-qty/price inputs and supplier/item autocompletes (same pattern as #21415)

Already merged to `development` via #21420; cherry-picked here (`4465149038`) since it was never carried to coop-prod (PR #21422 was closed unmerged; core fix confirmed still missing on coop-prod before this PR).

### 3. Refunded/retired lab items still showing in inpatient lab report (#21718)
`fetchLabBillItems()` only filtered `bi.bill.retired`, ignoring `bi.refunded` and `bi.retired` at the BillItem level, so refunded lab investigation items kept appearing in `inpatient_lab_investigation_item_list.xhtml`. Adds `bi.bill.cancelled = false`, `bi.retired = false`, `bi.refunded = false` to the JPQL.

Already merged to `development` via #21719 and to `coop-stg` via #21720; cherry-picked here (`408cc7f063`) since the coop-prod-specific hotfix PR #21721 was closed unmerged, and the underlying bug was confirmed still present on coop-prod before this PR.

Closes #22938, #21417, #21718

## Test plan
- [x] `mvn compile` passes clean
- [ ] Reviewer: confirm normal settle happy path unaffected
- [ ] Deploy to coop-prod and verify the abort path fires on a forced insufficient-stock scenario
- [ ] Verify PO request page: rapid double-submit no longer creates duplicate rows / zero-qty PBI
- [ ] Verify inpatient lab investigation report excludes refunded items

### DIFF
--- a/src/main/java/com/divudi/bean/inward/InwardReportControllerBht.java
+++ b/src/main/java/com/divudi/bean/inward/InwardReportControllerBht.java
@@ -466,12 +466,14 @@ public class InwardReportControllerBht implements Serializable {
     public List<BillItem> fetchLabBillItems(PatientEncounter pt, List<BillTypeAtomic> billTypesAtomics) {
         String jpql = "select bi "
                 + "from BillItem bi "
-                + "where bi.bill.retired = :ret "
+                + "where bi.bill.retired = false "
+                + "and bi.bill.cancelled = false "
+                + "and bi.retired = false "
+                + "and bi.refunded = false "
                 + "and bi.bill.billTypeAtomic in :billTypesAtomics "
                 + "and bi.bill.patientEncounter = :pe ";
 
         HashMap<String, Object> params = new HashMap<>();
-        params.put("ret", false);
         params.put("pe", pt);
         params.put("billTypesAtomics", billTypesAtomics);
 
--- a/src/main/java/com/divudi/bean/pharmacy/PurchaseOrderRequestController.java
+++ b/src/main/java/com/divudi/bean/pharmacy/PurchaseOrderRequestController.java
@@ -722,6 +722,7 @@ public class PurchaseOrderRequestController implements Serializable {
     public void saveBillComponent() {
         for (BillItem b : getBillItems()) {
             b.setBill(getCurrentBill());
+            resyncPharmaceuticalBillItemIfEmpty(b);
             if (b.getId() == null) {
                 getBillItemFacade().create(b);
             } else {
@@ -730,10 +731,32 @@ public class PurchaseOrderRequestController implements Serializable {
         }
     }
 
+    // A duplicate-line removal can leave the surviving BillItem with a lazily
+    // created all-zero PharmaceuticalBillItem while its BillItemFinanceDetails
+    // still holds the real quantities; downstream approval/GRN reads the PBI,
+    // so rebuild it from the finance details before persisting (issue #21417)
+    private void resyncPharmaceuticalBillItemIfEmpty(BillItem b) {
+        if (b == null || b.isRetired() || b.getBillItemFinanceDetails() == null) {
+            return;
+        }
+        PharmaceuticalBillItem pbi = b.getPharmaceuticalBillItem();
+        if (pbi.getQty() != 0 || pbi.getFreeQty() != 0) {
+            return;
+        }
+        BigDecimal qtyByUnits = b.getBillItemFinanceDetails().getQuantityByUnits();
+        BigDecimal freeQtyByUnits = b.getBillItemFinanceDetails().getFreeQuantityByUnits();
+        boolean financeDetailsHaveQty = (qtyByUnits != null && qtyByUnits.compareTo(BigDecimal.ZERO) > 0)
+                || (freeQtyByUnits != null && freeQtyByUnits.compareTo(BigDecimal.ZERO) > 0);
+        if (financeDetailsHaveQty) {
+            calculateLineValues(b);
+        }
+    }
+
     public void finalizeBillComponent() {
         getBillItems().removeIf(BillItem::isRetired);
         for (BillItem b : getBillItems()) {
             b.setBill(getCurrentBill());
+            resyncPharmaceuticalBillItemIfEmpty(b);
             BigDecimal qUnits = (b.getBillItemFinanceDetails() != null && b.getBillItemFinanceDetails().getQuantityByUnits() != null)
                     ? b.getBillItemFinanceDetails().getQuantityByUnits() : BigDecimal.ZERO;
             BigDecimal fqUnits = (b.getBillItemFinanceDetails() != null && b.getBillItemFinanceDetails().getFreeQuantityByUnits() != null)
@@ -797,7 +820,11 @@ public class PurchaseOrderRequestController implements Serializable {
         }
     }
 
-    private boolean saveRequestWithoutMessage() {
+    // synchronized: concurrent saves from the same session (Enter-key defaultCommand
+    // racing a button click) both saw billItem.id == null and persisted the same
+    // in-memory BillItem twice, creating duplicate rows sharing one
+    // BillItemFinanceDetails (issue #21417)
+    private synchronized boolean saveRequestWithoutMessage() {
         if (getCurrentBill().isChecked()) {
             JsfUtil.addErrorMessage("Cannot save a finalized bill");
             return false;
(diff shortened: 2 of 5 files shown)


buddhika75:
Updated the approach — same reasoning as the companion development PR (#22939#issuecomment-5289712971): a retire-based rollback couldn't cleanly unwind stock movements/request-qty tracking for earlier items in the same `settle()` call, so replaced it with prevention. `findInsufficientStockError()` aggregates every line's requested qty per Stock row across the whole `settle()` call and validates live DB stock can cover the combined total before any Bill/BillItem/PharmaceuticalBillItem is persisted.

This branch had two silent-skip points (older/diverged code) — both `isStockAvailable()` and `deductFromStock()`'s failure paths now call `logResidualStockRace()` for the narrow residual race case instead of silently continuing.

buddhika75:
Follow-up fix pushed (commit c78ce830a8): the receive side had its own crash bug, separate from the issue-side fix in this PR.

**Root cause (confirmed via coop production Payara log, hmis08, server.log_2026-08-13T15-43-35, 2026-08-13T15:34:04):**

\`\`\`
java.lang.NullPointerException
	at com.divudi.bean.pharmacy.TransferReceiveController.errorCheck(TransferReceiveController.java:959)
	at com.divudi.bean.pharmacy.TransferReceiveController.settle(TransferReceiveController.java:480)
\`\`\`

\`errorCheck()\` dereferenced \`pharmaceuticalBillItem.getStaffStock().getStock()\` without a null check. The Evion 400mg line's issue-side PBI had no Stock row (this PR's bug), so its \`staffStock\` was null when copied to the receive-side PBI, and \`errorCheck()\` crashed with an uncaught NPE while receiving TI/COOP/26/000488.

The crash aborted \`settle()\` mid-loop, leaving the request partially processed: 16 of 30 lines never actually moved stock from the porter to OPD Pharmacy despite the receive bill showing them as received, and no error was shown to the user.

Added a pre-loop check (\`findMissingStaffStockError()\`) that validates every line before any persistence starts, reporting a clear error naming the affected item instead of crashing.

**Data impact corrected 2026-08-14:** all 16 stuck items on the receive bill (14049028) via direct, audited SQL - porter stock zeroed, OPD Pharmacy stock credited/created, paired StockHistory rows inserted for traceability. Before/after values recorded in \`tmp/coop_stuck_receive_before_snapshot_2026-08-14.tsv\`.

Tracked in #22951.

Lakmedipro (approved):

PR #22985, opened 2026-08-15
feat(inward): show price-matrix Service Charge (Margin) on final bill

### LINKED ISSUE #22975: Display Matrix Value when settling the Inward Final Bill
RH Issue Tracker: 120

---

## **Describe the bug**

- There is a requirement for a **new column** to display the **applied Matrix value** when settling the Inward Final Bill.
- Currently, the final bill displays the total amount, but the **Price Matrix values are not displayed**. 
- Billing staff need to be able to identify the Matrix values and verify how the final **Net Total** has been calculated.

**Path:** `inward/inward_bill_final`

## **To Reproduce**

1. Admit a patient to Inward.
2. Add the required services, items, and other billable charges to the patient's BHT.
3. Generate the Inward Final Bill.
4. Navigate to the Inward Final Bill settlement page:
5. Review the billing details under "Charge Types".
6. Observe that the applied **Matrix values are not displayed separately**.

## **Expected behavior**

The Inward Final Bill settlement page should display:

- **Total**
- **Matrix Value**
- **Net Total**
- & etc.

The calculation should be clearly represented as:

**Net Total = Total + Price Matrix Value**

### Example

| Description        |         Amount |
| ------------------ | -------------: |
| Total              |     Rs. 10,000 |
| Matrix Value |        Rs. 500 |
| **Net Total**      | **Rs. 10,500** |

## **Screenshots**

Image

buddhika75:

This was run via `dev-issue-unattended` — the user was unreachable, so these judgment calls were made autonomously and are listed here for review:

1. **"Total already includes the matrix value" — didn't literally implement `Net Total = Total + Matrix Value`.** The issue's example table suggests Net Total should be Total plus the matrix value, but reading `BhtSummeryController`/`InwardBeanController` showed Total is already `Gross + Matrix Value` (confirmed via the `sum(grossValue+marginValue)` JPQL for services, and `PatientRoom.calculatedXxxCharge = base + marginXxxCharge` for room charges). Implementing the literal formula would have double-counted the matrix value into Net Total. Instead, the fix exposes the existing margin as its own read-only figure next to Total.
2. **Column/row label "Service Charge (Margin)"**, not "Matrix Value" as first drafted. Found that `inward_bill_intrim.xhtml` already shows this exact figure (`bhtSummeryController.marginTotal`) under the label "Service Charge (Margin)", gated by the `ShowServiceCharges` privilege — matched that existing convention instead of inventing a new term, and applied the same privilege gate to the new final-bill elements.
3. **Room/Maintain/MO/Nursing/Linen/Medical Care/Administration charges never had their margin summed at all**, not just displayed with a wrong number — this was a second, related gap found while investigating: `setGrossMarginVatBreakdown()` only summed margin for BillItem-backed and BillFee-backed charges, never for PatientRoom-backed ones, so `marginTotal` undercounted whenever a room-facility-category price matrix applied (not just on the final bill — the interim bill's existing "Service Charge (Margin)" figure was silently wrong too). Fixed by summing PatientRoom's margin fields into `ChargeItemTotal.margin` as well.
4. **Summed the room margin in Java from the already-recalculated `getPatientRooms()` list, not a JPQL query.** PatientRoom's `marginRoomCharge`/etc. fields are `@Transient` ("display-only, recomputed on every calculation" per the existing code comment) — a JPQL `SUM()` over them isn't valid. Discovered this only after an initial JPQL-based attempt compiled but would have failed at runtime; reverted and re-implemented in-memory.

## What changed

- `inward_bill_final.xhtml`: added a **Service Charge (Margin)** column to the Charge Types grid (next to Total) and a **Gross Total** / **Service Charge (Margin)** row pair to the Charges Overview panel (next to Total Charges), both gated by the `ShowServiceCharges` privilege.
- `BhtSummeryController.java`: added `sumRoomMatrixMargins()` and wired it into `setGrossMarginVatBreakdown()` so PatientRoom-backed charge types' matrix margin is included in `ChargeItemTotal.margin`/`marginTotal`, same as service/professional charges already were.

## Verification

Local Playwright E2E against a real (non-finalized) admission, after adding a temporary room-category price matrix row (10% margin, department/payment-method/room-category matched) via the app's own Price Matrix admin page:

- **Charges Overview**: Total Charges 43,590.79 = Gross Total 40,286.39 + Service Charge (Margin) 3,304.40.
- **Charge Types grid, Room Charges row**: Total 28,710.00, Service Charge (Margin) 2,610.00 — exactly 10% of the base, matching the configured margin.

Test matrix row retired afterward via the same admin page so it doesn't affect real billing.

### Screenshot (patient/BHT/room redacted)

Charges Overview and Charge Types showing Service Charge (Margin)

More detail: [wiki § Price Matrix Value (Service Charge Margin)](https://github.com/hmislk/hmis/wiki/Inpatient-Final-Bill-Generation#price-matrix-value-service-charge-margin).

Closes #22975

## Summary by CodeRabbit

* **New Features**
  * Payment overviews now display Gross Total and Service Charge (Margin) values based on access privileges.
  * Charge-type tables now include a Service Charge (Margin) column with explanatory tooltips.
  * Room-category margins are included in charge totals and margin calculations.

### DIFF
--- a/src/main/java/com/divudi/bean/inward/BhtSummeryController.java
+++ b/src/main/java/com/divudi/bean/inward/BhtSummeryController.java
@@ -4852,7 +4852,13 @@ public class BhtSummeryController implements Serializable {
      * breakdown, not just the net total. Services/investigations
      * (BillItem-backed) come from a bulk JPQL sum grouped by InwardChargeType;
      * Professional/Assisting fees (BillFee-backed, staff fee records) are
-     * summed from the lists already fetched for their respective tabs.
+     * summed from the lists already fetched for their respective tabs;
+     * Room/Maintain/MO/Nursing/Medical Care/Administration/Linen charges
+     * (PatientRoom-backed) are summed in-memory from the already-recalculated
+     * {@link #getPatientRooms()} list, because PatientRoom's marginXxxCharge
+     * fields are {@code @Transient} — display-only, recomputed on every
+     * calculation, never persisted — so a JPQL sum over them is not possible
+     * (issue #22975).
      */
     private void setGrossMarginVatBreakdown() {
         Map<InwardChargeType, double[]> serviceBreakdown = getInwardBean().calServiceBillItemsGrossMarginVatByInwardChargeTypeBulk(getPatientEncounter(), childPatientEncouters);
@@ -4872,6 +4878,21 @@ public class BhtSummeryController implements Serializable {
                 cit.setVat(values[2]);
             } else {
                 cit.setGross(cit.getTotal());
+            }
+        }
+
+        // Room/maintain/MO/nursing/medical-care/administration/linen charges are
+        // computed straight from PatientRoom (not BillItem), so they never show up
+        // in serviceBreakdown above. Their own price-matrix margin (issue #22975)
+        // lives on PatientRoom's marginXxxCharge fields instead, and Total already
+        // has it folded in the same way BillItem's does — pull it out here so the
+        // Charge Types grid can show it as its own Matrix Value column.
+        Map<InwardChargeType, Double> roomMargins = sumRoomMatrixMargins();
+        for (ChargeItemTotal cit : chargeItemTotals) {
+            Double roomMargin = roomMargins.get(cit.getInwardChargeType());
+            if (roomMargin != null && roomMargin != 0.0) {
+                cit.setMargin(roomMargin);
+                cit.setGross(cit.getTotal() - roomMargin);
             }
         }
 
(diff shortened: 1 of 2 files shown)


buddhika75:
Addressed both CodeRabbit findings in 1550e61:

1. **Stale/zero room margins after a live discount edit** — `sumRoomMatrixMargins()` now recomputes `calculateRoomCharge()`/`calculateMaintananceCharge()`/etc. unconditionally before summing, instead of relying on `getPatientRooms()`'s lazy-create (which skips recalculation once `patientRooms` is already non-null — as it is right after any of the seven per-charge-type discount listeners reload it straight from the DB). Verified these calculate methods never touch discount/adjusted fields, so this can't clobber a discount the user just edited.
2. **Overwriting instead of accumulating margin** — `cit.setMargin(roomMargin)` is now `cit.setMargin(cit.getMargin() + roomMargin)`, so a charge type that combines a room-backed total with a service-backed one no longer loses the service-side margin.

Re-verified locally via Playwright against the same test admission — Charges Overview and Charge Types grid figures are unchanged (this repo's local data doesn't happen to exercise either edge case), confirming both fixes are no-ops on the already-tested path.

buddhika75:
@coderabbitai review

Issue #20729, opened 2026-05-14
perf: audit which StockHistory reports need access to archived rows


buddhika75:
## Follow-up to #20726

After landing the StockHistory archive job, audit which reports / API endpoints query `StockHistory` and decide whether any of them must also include archived rows.

### Background

Issue #20726 introduces `STOCKHISTORYARCHIVE` and moves rows older than the configured retention window (default: 2 years) out of `STOCKHISTORY`. Existing read paths still query only `StockHistory`, so they will silently lose visibility of archived data after the first archival run.

Most pharmacy operational reports (daily stock value, current stock at date, last N months trend, etc.) operate on recent data and are unaffected. But there may be edge cases that need long-range history.

### Tasks

- [ ] Inventory all JPQL / native SQL queries against `StockHistory` (services, facades, REST endpoints, scheduled tasks)
- [ ] For each, decide: operational only (no change), or also needs archive (add a query mode)
- [ ] For queries that need archive: pick one of:
  - Add a UNION-style helper on `StockHistoryFacade` (e.g. `findIncludingArchive(...)`) that queries both tables
  - Add a dedicated archive-only browse/search page for ad-hoc lookups
  - Both
- [ ] Update affected screens with an "include archived" toggle if user-facing

### Acceptance

Documented audit results in `developer_docs/pharmacy/stock-history-archive-read-paths.md` covering each query identified, the decision taken, and the migration applied (if any).

### Related

- #20726 (parent — archive job + admin page)
- #20724 (sibling — ItemBatch archive; same read-side audit will likely be needed)

PR #23025, opened 2026-08-16
fix(reports): stop dropping zero/negative-qty items from COGS Opening/Closing Stock

### LINKED ISSUE #23018: COGS report: Opening/Closing Stock silently drops zero/negative-qty items, causing phantom variance that masks itself over wider date ranges
## Summary

The Cost of Goods Sold (COGS) report's Opening Stock and Closing Stock rows silently exclude any item whose running stock quantity is zero or negative as of the snapshot date. This produces a phantom "Variance" row on date ranges that happen to end while an item is in a negative/zero stock state, and the variance can appear to "disappear" when the date range is widened past the point where the item's stock recovers to positive — even though nothing was actually fixed.

## Where

`src/main/java/com/divudi/bean/report/PharmacyReportController.java`, method `calculateStockTotals(Date date)` (~line 12310-12435).

The query has:

```java
jpql.append("GROUP BY sh4.itemBatch.item.id ")
        .append("HAVING SUM(sh4.stockQty) > 0");
```

Immediately below this (lines ~12407-12413) there is a comment that already documents this exact defect and says it must not be there:

> // NOTE: do NOT filter batches/items by SUM(stockQty) > 0 here. This snapshot
> // must include every batch's latest quantity as-of `date`, positive, zero, or
> // negative (oversold/backorder). Excluding negative-quantity items understates
> // Opening/Closing Stock only on the snapshot where the item's running total
> // happens to be negative, breaking the Opening + movements = Closing identity
> // and producing a phantom COGS variance the moment that item's total crosses
> // zero within the report window (issue: COGS report variance, 2026-08-11).

The comment describing the fix was added, but the actual `HAVING SUM(sh4.stockQty) > 0` filter (and its consignment-item counterpart `AND SUM(sh4.stockQty) <= 0`) was never removed, so the bug is still live.

## Reproduction (verified via DB, Ruhunu new infra, 2026-08-16)

Department: Matara - Pharmacy (id 3193897).

- Report for **10 Jul 2026** shows Variance: Purchase -670.32, Cost -670.32, Retail -782.26.
- Widening the range to **10 Jul 2026 → 28 Jul 2026** makes the variance disappear.

Root cause traced to two item-batches sitting at physically negative stock on 10 Jul 2026 (known fallout of an earlier "cancel return" mass-click data-integrity incident at Matara Pharmacy on 10 Jul, tracked separately):

- Astrocort Cream 1% 15G, batch 5164564: qty -2 → -564.00 purchase / -660.00 retail
- Hexilon 16Mg Tablet, batch 4609211: qty -2 → -106.32 purchase / -122.26 retail

Sum: -670.32 purchase, -782.26 retail — exactly matches the report's Variance row for 10 Jul.

By 28 Jul 2026, Hexilon's batch had been restocked to +49 (crosses back above 0, so it's re-included), removing 106.32/122.26 of the mismatch. Astrocort sits at exactly 0 on both the 10th and 28th, so it's excluded consistently at both ends of the range and its exclusion cancels out in the Opening-vs-Closing subtraction. The variance is masked by the range boundary, not resolved.

## Fix

Remove the `HAVING SUM(sh4.stockQty) > 0` filter (and the paired `isConsignmentItem()` `<= 0` branch) from `calculateStockTotals()`, per the existing adjacent comment. Opening/Closing Stock snapshots should include every item-batch's latest quantity as-of the snapshot date regardless of sign, exactly like every other COGS row already does.

## Note

This does not fix the underlying negative stock at Matara Pharmacy (Astrocort batch 5164564, Hexilon batch 4609211) — that's a separate data-correction task tied to the 10-July incident. This issue is scoped to the report calculation bug only.

buddhika75:
## Hotfix

Ports the COGS report fix from #23019 (already merged to `development`, deployed and verified on `rh-stg` and `rh-stg-migrated`) to `ruhunu-prod-migrated`.

## Summary
- `calculateStockTotals()` in `PharmacyReportController` had a `HAVING SUM(sh4.stockQty) > 0` filter that silently excluded any item-batch sitting at zero/negative quantity as of the report's snapshot date from both Opening Stock and Closing Stock rows.
- This produced a phantom "Variance" row whenever a date range's boundary fell while an item was oversold/backordered, and made the variance appear to self-resolve when the range was widened past the date the item's stock recovered — without anything actually being fixed.
- Fix: remove the item-level `HAVING`/consignment-branch filter entirely. Opening/Closing Stock now include every item-batch's latest snapshot quantity regardless of sign, consistent with every other COGS row.

## Root cause (reproduced against Ruhunu Matara Pharmacy data, dept 3193897)
Report for 10 Jul 2026 showed Variance: Purchase -670.32 / Cost -670.32 / Retail -782.26, caused by two batches sitting at negative stock that day (Astrocort batch 5164564, Hexilon batch 4609211) - fallout of an unrelated, separately-tracked "cancel return" data-integrity incident at Matara Pharmacy. Removing the filter makes Closing Stock match Calculated Closing Stock Value exactly (Variance 0), verified live on `rh-stg-migrated`.

## Note
This does not touch the underlying negative stock at Matara Pharmacy - that data correction is tracked separately, still awaiting a physical stock count.

Closes #23018
Ports #23019 to `ruhunu-prod-migrated`

## Test plan
- [x] `mvn compile` succeeds
- [x] Fix already verified live on `rh-stg-migrated` (COGS report for 10 Jul 2026, Matara Pharmacy: Variance 0.00)
- [x] Diff scoped to the single method, no unrelated changes

### DIFF
--- a/src/main/java/com/divudi/bean/common/OpdBatchBillCancellationController.java
+++ b/src/main/java/com/divudi/bean/common/OpdBatchBillCancellationController.java
@@ -3034,7 +3034,6 @@ public class OpdBatchBillCancellationController implements Serializable, Control
 
         originalBill.setCancelled(true);
         originalBill.setCancelledBill(individualCancelltionBill);
-        originalBill.setBillItems(null);
         billService.saveBill(originalBill);
     }
 
@@ -3179,7 +3178,6 @@ public class OpdBatchBillCancellationController implements Serializable, Control
 
         originalBill.setCancelled(true);
         originalBill.setCancelledBill(individualCancelltionBill);
-        originalBill.setBillItems(null);
         billService.saveBill(originalBill);
     }
 
--- a/src/main/java/com/divudi/bean/common/OpdBatchBillCancellationController.java
+++ b/src/main/java/com/divudi/bean/common/OpdBatchBillCancellationController.java
@@ -3032,9 +3032,9 @@ public class OpdBatchBillCancellationController implements Serializable, Control
         }
         billService.saveBill(individualCancelltionBill);
 
-        originalBill.setBillItems(billService.fetchBillItems(originalBill));
         originalBill.setCancelled(true);
         originalBill.setCancelledBill(individualCancelltionBill);
+        originalBill.setBillItems(null);
         billService.saveBill(originalBill);
     }
 
@@ -3177,9 +3177,9 @@ public class OpdBatchBillCancellationController implements Serializable, Control
         }
         billService.saveBill(individualCancelltionBill);
 
-        originalBill.setBillItems(billService.fetchBillItems(originalBill));
         originalBill.setCancelled(true);
         originalBill.setCancelledBill(individualCancelltionBill);
+        originalBill.setBillItems(null);
         billService.saveBill(originalBill);
     }
 
--- a/src/main/java/com/divudi/bean/report/PharmacyReportController.java
+++ b/src/main/java/com/divudi/bean/report/PharmacyReportController.java
@@ -11089,13 +11089,16 @@ public class PharmacyReportController implements Serializable {
             }
 
             // Consignment filter: Use item stock quantity (not batch qty)
+            // NOTE: hide zero-stock items (normal, expected), but never hide negative
+            // quantities - a negative net qty means oversold/backorder data that must
+            // stay visible instead of being silently dropped from the report (issue #23026).
             double itemQty = row.getStockQty() != null ? row.getStockQty() : 0.0;
             if (isConsignmentItem()) {
                 if (itemQty > 0) {
                     continue;
                 }
             } else {
-                if (itemQty <= 0) {
+                if (itemQty == 0) {
                     continue;
                 }
             }
@@ -11315,13 +11318,16 @@ public class PharmacyReportController implements Serializable {
             }
 
             // Consignment filter (unchanged - stays in Java)
+            // NOTE: hide zero-stock batches (normal, expected), but never hide negative
+            // quantities - a negative batch qty means oversold/backorder data that must
+            // stay visible instead of being silently dropped from the report (issue #23026).
             double batchQty = row.getStockQty() != null ? row.getStockQty() : 0.0;
             if (isConsignmentItem()) {
                 if (batchQty > 0) {
                     continue;
                 }
             } else {
-                if (batchQty <= 0) {
+                if (batchQty == 0) {
                     continue;
                 }
             }
--- a/src/main/java/com/divudi/bean/report/PharmacyReportController.java
+++ b/src/main/java/com/divudi/bean/report/PharmacyReportController.java
@@ -11746,46 +11746,13 @@ public class PharmacyReportController implements Serializable {
                 jpql.append("AND sh.item.departmentType IN :departmentTypes ");
             }
 
-            // Group by item and filter positive quantities
-            jpql.append("AND sh.itemBatch.item.id IN (")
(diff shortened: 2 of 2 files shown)


damithdeshan98 (approved):
Approved

buddhika75:
Added a second commit to this PR: also fixes the Closing/Opening Stock drill-down report (`closing_stock_report.xhtml`, reached by clicking the "Closing Stock"/"Opening Stock" row in the COGS report), which had the same class of bug - filtering out rows with net quantity `<= 0` instead of `== 0`, silently dropping negative-quantity items. See #23028 for details.

Note: this branch's `processClosingStockForItemReport`/`processClosingStockForBatchReport` don't have the `includeArchived` blocks that exist on `development` (that feature - #20729 - hasn't been ported here), so only the live-data filter conditions were changed; nothing archive-related was introduced.

buddhika75:
Added a second fix in this PR while the deploy window is open tonight:

**fix(opd): stop cross-bill BillFee corruption on batch bill cancellation**

Root-caused during the Ruhunu Labeema Daily Return vs QB Report investigation. `Bill.billItems` is mapped `orphanRemoval=true`; `cancelSingleBillWhenCancellingOpdBatchBill()` and its package-bill twin re-fetched and re-set `originalBill`'s items right before the final `edit()` merge that marks the bill cancelled. In the `@SessionScoped` cancellation controller, two similar OPD batch cancellations run back-to-back in one session can leave that re-fetched collection reflecting stale JPA state, so the orphanRemoval cascade re-parents the new cancellation's `BillFee` rows onto the *wrong* bill's `BillItem`. Payment-based reports (Daily Return, Cashier Summary) are unaffected since they read a different table — only item/fee-based reports (QB Report) show the corruption.

Confirmed via git history this exact path was never touched by a prior fix (the established pattern from the pharmacy module's `orphanRemoval` fixes, e.g. `1e263fc3ff`, was never ported to this OPD controller). Live DB evidence: 28 bills across 6 OPD departments corrupted between 29.01.2025 and 15.02.2026 — dormant since only because no OPD batch cancellations have run since, not because of any prior fix.

Fix: `setBillItems(null)` immediately before that merge, so EclipseLink skips the collection instead of treating a stale snapshot as authoritative — same guard already proven in the pharmacy module.

Verified: `mvn compile` succeeds, diff scoped to the two affected merge call sites only.

Will also port this same fix to `development` in a separate PR (this branch targets `ruhunu-prod-migrated` only).

buddhika75:
Added an additional commit to this branch (unrelated to the COGS report fix this PR was originally about, but this branch is the current path to \`ruhunu-prod-migrated\`):

This branch already carried commit \`1a4485d331\` ("fix(opd): stop cross-bill BillFee corruption on batch bill cancellation"), a port of PR #23040's fix. That fix's \`originalBill.setBillItems(null)\` turned out to crash OPD batch cancellation with a foreign-key violation on \`DELETE FROM BILLITEM\` whenever the original bill has a \`BillComponent\` row (i.e. essentially all normal OPD bills) — because \`Bill.billItems\` is \`orphanRemoval=true\`, so setting the collection to \`null\` schedules every previously-attached \`BillItem\` for delete instead of being a no-op, and \`BillComponent.billItem\` has no cascade back to that row.

Root-caused and fixed on \`development\` in #23048, and ported here with the identical two-line change: remove the \`setBillItems(...)\` call entirely rather than setting it to \`null\`. Verified via Playwright + DB against local \`coop\`: a same-patient repeat-cancellation (the original real-world trigger) now completes without error, with zero cross-bill \`BillFee\` corruption and the original bill's own \`BillItem\`/\`BillComponent\` left untouched.

PR #23057, opened 2026-08-16
docs(demonstrate-issues): handle content-free capture cues and forward-looking narration

### LINKED ISSUE #23056: demonstrate-issues skill: handle content-free capture cues and forward-looking narration
## Summary
Two process gaps in the `demonstrate-issues` skill (`.claude/skills/demonstrate-issues/SKILL.md`) surfaced during a live multi-bug demonstration session on 2026-08-16, both around ambiguous or forward-looking user cues during the demonstration loop (§3).

## Gap 1 — No guidance for content-free capture cues
When the user's cue to capture carries no description at all (e.g. just "capture", or "check playwrite"), the skill currently gives no instruction on how to proceed. In practice this meant capturing evidence speculatively, then asking the user "what's the actual issue you want recorded?" — adding an extra round-trip before the demo could be confirmed.

**Suggested fix:** add explicit guidance to §3 (Demonstration loop) that when a capture cue carries no description, the assistant should ask for the one-line description either before capturing or immediately alongside the capture, rather than capturing first and asking afterward.

## Gap 2 — No way to downgrade a captured demo to "context only"
During this session, a capture was recorded as a confirmed demo (post-settle screenshot + console error), but the user's very next message revealed it wasn't meant as a bug report at all — it was scene-setting for a later demo ("I am going to show that if I order X, the issue should propose Y... no error in this page yet, just want to gather facts for main issue"). The skill has no affordance for the user to mark a prior capture as "context only, not a confirmed bug," so this had to be caught and re-confirmed with the user during the investigation phase (§5) instead of being resolved cleanly at demonstration time.

**Suggested fix:** add explicit guidance to §3 for handling forward-looking/narrative cues ("I'm going to show...") distinctly from firm bug cues, and document a lightweight way for the user to say "that last capture was just context, drop it" that the assistant should recognize and act on immediately rather than carrying the ambiguity into the investigation phase.

## Environment
- Branch/commit: `development` @ `6706e89f`
- Skill file: `.claude/skills/demonstrate-issues/SKILL.md`

## Evidence
No screenshot — this is a process/documentation issue in the skill definition itself, not a UI bug. See the session transcript for the concrete exchange that prompted Gap 2 (the "no error in this page yet, just want to gather facts" message immediately following a confirmed capture).

buddhika75:
## Summary
Fixes #23056 — adds guidance to the `demonstrate-issues` skill for two process gaps found during a live multi-bug demonstration session.

- **Where to place the new guidance**: added as two new subsections under existing §3 "Demonstration loop" rather than a new top-level section, since both gaps are specifically about that phase (capture-cue handling), not the investigation or filing phases.
- **Scope of the "drop a prior capture" mechanism**: kept it lightweight (acknowledge + exclude from later phases) rather than introducing a new state/numbering scheme for demos, to match the skill's existing low-ceremony tone.

## Changes
- `.claude/skills/demonstrate-issues/SKILL.md` — §3 now instructs:
  1. Asking for a one-line description up front when a capture cue carries none (e.g. "capture" alone), instead of capturing then asking afterward.
  2. Distinguishing forward-looking narration ("I'm about to show you X") from a firm bug cue, and explicitly dropping/acknowledging a prior capture if the user's next message reveals it wasn't meant as a bug report.

## Verification
Doc-only change to a skill markdown file — no build/deploy/Playwright verification applicable (same as any JSF-only/no-Java change per this repo's testing rules).

## Summary by CodeRabbit

* **Documentation**
  * Updated demonstration guidance to require a description before capturing cues.
  * Clarified the difference between planned narration and actual bug reports.
  * Added instructions for discarding non-bug demonstrations before investigation or filing.

### DIFF
--- a/.claude/skills/demonstrate-issues/SKILL.md
+++ b/.claude/skills/demonstrate-issues/SKILL.md
@@ -112,9 +112,9 @@ picture of each one is understood.
 
 If the cue to capture carries no description at all (e.g. just "capture", or
 "check playwrite"), don't capture speculatively and ask afterward what it
-was for — ask for the one-line description first (or in the same turn as the
-capture), so each recorded demo has its description attached from the start
-instead of needing a follow-up round-trip.
+was for — ask for the one-line description. If the user provides it in the
+same turn, capture the demo; otherwise wait and do not capture until the
+description is available.
 
 ### Forward-looking narration vs a firm bug cue
 
@@ -123,9 +123,10 @@ happened yet) from "see this, X is broken" (the actual cue). Narration that
 describes what's coming next is not itself a capture cue — wait for the
 concrete bug before recording anything.
 
-If a capture already happened and the user's next message reveals it wasn't
+If a capture already happened and the user later clarifies that demo #N was not
 meant as a bug report (e.g. "no error in this page yet, just gathering
-facts"), treat that as an explicit instruction to drop the prior capture:
+facts"), treat that as an explicit instruction to drop the prior capture —
+whether that clarification arrives in the very next message or a later one:
 acknowledge it ("Dropping demo #N — noted as context only") and exclude it
 from the investigation/filing phases. Don't carry the ambiguity forward and
 make the user re-resolve it during investigation.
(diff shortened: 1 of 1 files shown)


buddhika75:
Applied both wording fixes as suggested. The SkillSpector 'credential access' flag on the same lines is a false positive — that's the generic `C:\Credentials\credentials.txt` placeholder path from CLAUDE.md's own credential-handling convention (never inline real values, point to external storage), not a real credential read.

Issue #21974, opened 2026-07-09
Channel Session Report Does Not Display Room Number in "Room No" Column


chathushkiupulini:
**Describe the bug**
The report is generated, and the Room No column is displayed. However, the column does not contain any room number values for the consultant's channeling sessions.

**To Reproduce**
File path:  channel/doctor_card.xhtml

1. Navigate to Channeling.
2. Open the Doctor Card.
3. Navigate to the Channel Session Report page.
4 Enter the required filters:
 Date
Specialty
Consultant
6. Click the Generate Report button.

**Expected behavior**
The Room No column should display the correct room number assigned to each consultant's channeling session in the generated report.

**Screenshots**

<img width="1917" height="968" alt="Image" src="https://github.com/user-attachments/assets/3a4588be-e434-4207-835c-07c06d27b6b7" />

<img width="1240" height="857" alt="Image" src="https://github.com/user-attachments/assets/ea1e46ec-863b-4d2e-9438-795cab56ff0b" />

PR #23082, opened 2026-08-17
21974 Channel Session Report Does Not Display Room Number in "Room No" Column

SheharaHeshan:
## Summary by CodeRabbit

* **Bug Fixes**
  * Doctor card reports now display the correct room number without an extraneous value.
  * Session information and booking counts are updated more accurately for active channel appointments.

### DIFF
--- a/src/main/java/com/divudi/bean/channel/ChannelReportController.java
+++ b/src/main/java/com/divudi/bean/channel/ChannelReportController.java
@@ -1294,8 +1294,6 @@ public class ChannelReportController implements Serializable {
         for (SessionInstance s : sessioninstances) {
             if (s.isCancelled()) {
                 canceledSessionInstances.add(s);
-            }else{
-            channelService.fillBillSessionsAndUpdateBookingsCountInSessionInstance(s);
             }
         }
 
--- a/src/main/java/com/divudi/bean/channel/ChannelReportController.java
+++ b/src/main/java/com/divudi/bean/channel/ChannelReportController.java
@@ -1294,6 +1294,8 @@ public class ChannelReportController implements Serializable {
         for (SessionInstance s : sessioninstances) {
             if (s.isCancelled()) {
                 canceledSessionInstances.add(s);
+            }else{
+            channelService.fillBillSessionsAndUpdateBookingsCountInSessionInstance(s);
             }
         }
 
--- a/src/main/webapp/channel/doctor_card.xhtml
+++ b/src/main/webapp/channel/doctor_card.xhtml
@@ -126,11 +126,11 @@
                                 </p:outputLabel>
                             </p:column>
                             <p:column style="padding: 6px">
-                                <f:facet name="header" >
-                                    <p:outputLabel value="Room No" />
-                                    0   </f:facet>
-
-                            </p:column>
+                               <f:facet name="header" >
+                                  <p:outputLabel value="Room No" />
+                              </f:facet>
+                              <p:outputLabel value="#{i.roomNo}" />
+                           </p:column>
                         </p:dataTable>
                     </p:panel>
 


src/main/java/com/divudi/bean/channel/ChannelReportController.java
```diff
@@ -1294,6 +1294,8 @@ public void fillSessionsForChannelDoctorCard() {
         for (SessionInstance s : sessioninstances) {
             if (s.isCancelled()) {
                 canceledSessionInstances.add(s);
+            }else{
+            channelService.fillBillSessionsAndUpdateBookingsCountInSessionInstance(s);
```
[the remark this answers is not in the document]
SheharaHeshan: no issue both the room number and count columns will display the correct and updated values with the loop based chnage.

buddhika75, line 1298: **[Correctness] Race condition / lost update from calling a mutating booking-service method in a report-view loop.**

`channelService.fillBillSessionsAndUpdateBookingsCountInSessionInstance(s)` doesn't just compute a display count — it recomputes and persists `SessionInstance.nextAvailableAppointmentNumber` (`ChannelService.java` ~line 1421-1426, ends in `sessionInstanceFacade.edit(session)`), a field with no `@Version`/optimistic-locking that real booking code (`ChannelService.createBillSession`, `BookingControllerViewScope`) treats as authoritative for assigning live appointment numbers.

The method's other two call sites (`ChannelService.java:2549`, `3135`) each persist a single `SessionInstance` in-transaction immediately after an actual booking/payment mutation on that instance. This PR is the first to call it in a loop purely to populate a report's "Count" column, with no underlying booking state change.

**Failure scenario:** Staff runs this Doctor Card 'Fill' report over a date range while a patient is concurrently being booked (walk-in or online) against one of the sessions in range. The report's read-then-`edit()` can race with the concurrent booking transaction's write to the same unversioned field — a lost update — leading to a duplicate or incorrect appointment number being handed out next.

Related to the CodeRabbit performance note above on this same line, but this is a separate correctness-class concern (data race), not just extra query/write cost.

src/main/java/com/divudi/bean/channel/ChannelReportController.java
```diff
@@ -1294,6 +1294,8 @@ public void fillSessionsForChannelDoctorCard() {
         for (SessionInstance s : sessioninstances) {
             if (s.isCancelled()) {
                 canceledSessionInstances.add(s);
+            }else{
```
buddhika75, line 1297: **[Performance] No date-range cap, and doctor/speciality filters are optional — the per-row query+write loop below can run unbounded.**

Unlike the sibling method `fetchChannelIncomeByUser` (~line 1219-1225 in this file), `fillSessionsForChannelDoctorCard` has no fromDate/toDate range validation, and `speciality`/`staff` are optional filters. Combined with the new per-`SessionInstance` query+write added in this PR (line 1298 below), a user can leave speciality/staff blank and pick a multi-month range on `doctor_card.xhtml` — whose table has no paginator and whose date pickers have no min/max constraint — causing this non-AJAX request to issue one JPQL query plus one UPDATE per session synchronously, risking a timeout or DB-connection-pool exhaustion on a hospital-scale dataset.

buddhika75:
## 🚦 Merge Gate: BLOCKED — code review found regressions

Two correctness/performance findings posted as inline comments above (`ChannelReportController.java:1297` and `:1298`):
1. **Race condition / lost update** — the new report code calls a mutating booking-service method that persists an unversioned `nextAvailableAppointmentNumber` field purely to populate a display count, racing against real concurrent bookings.
2. **Unbounded N+1 risk** — no date-range cap and optional doctor/speciality filters mean a wide-range report triggers one query + one write per session synchronously.

Merge-gate stopped here; Phase 2/3 (build, E2E, baselines) did not run — **no screenshots/E2E evidence exist for this PR**, only these code-level findings.

Re-run `/merge-gate` on this PR once addressed.

src/main/java/com/divudi/bean/channel/ChannelReportController.java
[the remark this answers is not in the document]
buddhika75: Fixed in e7aaf26680: removed the fillBillSessionsAndUpdateBookingsCountInSessionInstance(s) call from this loop entirely rather than batching it. It was unrelated to the Room No fix this PR targets (issue #21974) — the Room No column now renders directly from the persisted SessionInstance.roomNo field via #{i.roomNo} in doctor_card.xhtml, no service call needed. Verified live: Room No renders correctly (e.g. "25") for sessions with a set room number, and blank for sessions without one.

buddhika75: Fixed in e7aaf26680: removed the call entirely (see reply on the sibling performance thread on this same line). This eliminates the race/lost-update risk along with the N+1 cost, since the mutating booking-service call no longer runs from this report-view loop at all.

src/main/java/com/divudi/bean/channel/ChannelReportController.java
buddhika75: Fixed in e7aaf26680: removed the per-session query+write call from this loop entirely, so the unbounded-date-range concern no longer applies here. The date-range validation gap on fillSessionsForChannelDoctorCard itself is pre-existing and out of scope for this Room No fix (#21974) — happy to file a follow-up issue if wanted.

PR #23096, opened 2026-08-18
fix(collecting-centre): charge CC settlement amount from itemized bills, not ledger delta

### LINKED ISSUE #23095: CC settlement charges wrong amount from ledger delta instead of itemized bills
## Problem

`processCollectingCentrePayment()` calculated the amount charged to settle a Collecting Centre from the ledger's overall balance movement across the selected date range, independent of which specific bills were being marked paid.

Selecting a date range that started later than the true unpaid backlog would silently roll the backlog's value into the charged amount while leaving the older bills flagged unpaid, and a stale/unset `paymentDone` flag let the same range be re-settled and re-charged later.

This produced a ~9-10x cash/bill-value mismatch on two collecting centres and led to a wrongful "duplicate payment" cancellation that reinstated hundreds of thousands of rupees of phantom debt. Found and fixed in Ruhunu Prod (2026-08-17).

A related bug: `getAllHistoryFromPaymentBill()` bound its `AgentHistory` query with `BETWEEN ccPaymentBill.getFromDate() AND ccPaymentBill.getToDate()`. Payment bills created before `fromDate`/`toDate` existed on the `Bill` entity have both fields NULL, so `BETWEEN NULL AND NULL` silently matched zero rows — `cancelPaymentBill()` would then reverse the bill's cash/BillItems correctly but leave the underlying `AgentHistory` rows permanently stuck with `paymentDone=true`.

## Fix

- The amount charged now always equals the sum of the itemized bills being settled (`totalCCAmount`), not the ledger delta, and a new guard blocks settling a range that leaves older unpaid bills behind it.
- `getAllHistoryFromPaymentBill()` now falls back to the min/max `createdAt` of the bills actually referenced by the payment bill's BillItems when `fromDate`/`toDate` are null, so cancelling one of these legacy bills clears `paymentDone` correctly instead of no-op'ing.
- Also switched the CC Payment Bill search/export to a lightweight `CollectingCentrePaymentBillDTO` (instead of loading full Bill entities/relationships) and added From/To date columns to the search table.

## Impact

Financial correctness bug affecting Collecting Centre settlement cash reconciliation. Already deployed as a hotfix to Ruhunu Prod; this issue tracks merging it into `development`.

damithdeshan98:
## Summary
- `processCollectingCentrePayment()` now charges the Collecting Centre settlement amount as the sum of the itemized bills being settled (`totalCCAmount`) instead of the ledger's overall balance delta across the date range, and guards against settling a range that leaves older unpaid bills behind it.
- `getAllHistoryFromPaymentBill()` falls back to the min/max `createdAt` of the bills referenced by the payment bill's BillItems when `fromDate`/`toDate` are null, so cancelling legacy (pre-date-field) payment bills correctly clears `paymentDone` instead of no-op'ing.
- CC Payment Bill search/export switched to a lightweight `CollectingCentrePaymentBillDTO` instead of loading full Bill entities/relationships, with From/To date columns added to the search table.

## Why
The ledger-delta calculation produced a ~9-10x cash/bill-value mismatch on two collecting centres in Ruhunu Prod and led to a wrongful "duplicate payment" cancellation that reinstated hundreds of thousands of rupees of phantom debt. Hotfixed directly to Ruhunu Prod on 2026-08-17; this PR merges that fix into `development`.

Closes #23095

## Test plan
- [x] Verified in Ruhunu Prod: settlement amount now matches itemized bill totals for the affected collecting centres
- [x] Verified cancelling a legacy null-date payment bill clears `paymentDone` on its underlying AgentHistory rows
- [ ] Regression check of CC Payment Bill search/export table (DTO-backed) in development environment

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,17 @@
+{
+  "permissions": {
+    "additionalDirectories": [
+      "C:\\Development\\hmis.wiki"
+    ],
+    "allow": [
+      "Bash(mysql *)",
+      "Bash(curl -s *)",
+      "Bash(curl -sS *)",
+      "Bash(curl -s\"*)",
+      "Bash(ping -c *)",
+      "Bash(ss -tnlp*)",
+      "Bash(/home/buddhika/payara/bin/asadmin list*)",
+      "Bash(/home/buddhika/payara/bin/asadmin ping-connection-pool*)"
+    ]
+  }
+}
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -7,7 +7,6 @@
       "Bash(mysql *)",
       "Bash(curl -s *)",
       "Bash(curl -sS *)",
-      "Bash(curl -s\"*)",
       "Bash(ping -c *)",
       "Bash(ss -tnlp*)",
       "Bash(/home/buddhika/payara/bin/asadmin list*)",
--- a/developer_docs/database/dual-version-id-allocator-separation.md
+++ b/developer_docs/database/dual-version-id-allocator-separation.md
@@ -1,0 +1,134 @@
+# Dual-Version ID Allocator Separation (Temporary)
+
+**Status:** TEMPORARY measure — remove once every production deployment that shares a
+database with a newer build has been upgraded to `GenerationType.IDENTITY`.
+
+**Added for:** Ruhunu, June 2026. The Ruhunu production build and the development
+build run side by side against the same main and audit databases.
+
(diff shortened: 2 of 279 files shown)


damithdeshan98:
Superseded by #23097 — this branch was rebuilt on `origin/development` (was previously based on old master history, which pulled in ~90 unrelated commits into the diff). GitHub auto-closed this PR after the force-push and it can't be reopened.

Issue #23139, opened 2026-08-19
Add Stock Take Variance Detailed Report (before/after adjustment quantities & values)


buddhika75:
## Summary
Add a new "Stock Take Variance Detailed Report" page, linked from the existing Stock Take Variance Report (`pharmacy_stock_take_variance.xhtml`) via a new "Variance Detailed Report" button, showing per-batch quantity and value (cost/retail/purchase) **before and after the stock adjustment was posted**, with column totals as a table footer. **No changes to the existing Variance page's current behavior** — only an additional button is added to it.

## Environment
- Branch/commit: `development` @ `841356bd98`
- Department: OPD Pharmacy, Galle Co Operative Hospital
- Entry point: Pharmacy → Stock Take → Manage Stock Takings → Variance (on a completed stock take)

## Background / Data source
The existing Variance report (`PharmacyStockTakeController.prepareVarianceRows()`) shows the *snapshot* quantity and a *sum of variances* derived from the uploaded physical count. That is not the same as the actual before/after stock values written when the adjustment was posted.

The authoritative before/after values already exist on the **adjustment bill** — the `forwardReferenceBill` of the approved `PharmacyPhysicalCountBill` (built in `PharmacyStockTakeController` around lines 3623-3663). For each adjustment `BillItem`, its `PharmaceuticalBillItem` carries:
- `beforeAdjustmentValue` — stock quantity immediately before the adjustment (`stock.getStock()` at approval time)
- `afterAdjustmentValue` — target quantity after the adjustment (the physical count value)

These are set only for lines where `variance != 0` (line 3628 skips zero-variance items when building the adjustment bill), so — same as the existing Variance report — unchanged items won't appear in the new report either.

## Requirements

**New page**: `pharmacy_stock_take_variance_detailed.xhtml`, following the same read-only, scalar-JPQL-projection pattern as `prepareVarianceRows()` (no entity mutation, no native SQL — JPQL only per project rules).

**New controller method(s)**: e.g. `gotoViewVarianceDetailed(Bill snapshotBill)` navigating to the new page, and a `prepareVarianceDetailedRows()` that loads the adjustment bill items (via the snapshot bill → physical count bill(s) → `forwardReferenceBill`) with their `beforeAdjustmentValue`/`afterAdjustmentValue` and joined batch/rate data (purchase rate, retail rate, cost rate — same joins as the existing query: `PharmaceuticalBillItem` → `ItemBatch` → `Item`).

**Columns** (per batch line):
[...]

PR #23140, opened 2026-08-19
feat(pharmacy): add Stock Take Variance Detailed Report (before/after adjustment)

buddhika75:

This PR was developed unattended via `dev-issue-unattended`. Judgment calls made without a live check-in:

- **Data source**: sourced before/after quantities from the posted **adjustment bill's** `PharmaceuticalBillItem.beforeAdjustmentValue`/`afterAdjustmentValue` (the value actually written to `Stock` at approval time), not from the existing Variance Report's snapshot-qty + variance-sum rows. This was already the plan agreed with the user before this PR (see issue body), re-confirmed by reading `PharmacyStockTakeController.doApprovalLogic()` where these fields are set.
- **Rate columns reflect a pre-existing data gap, left as-is**: while verifying, I found that `PharmacyStockAdjustmentBill` items never get `purchaseRate`/`retailRate`/`costRate` set consistently — historically `purchaseRate`/`retailRate` were populated by an older code path but `costRate` is `0` for all adjustment lines in the data I tested (3,110/3,110 rows). This report faithfully displays whatever is in the adjustment line, so "Value Before/After at Cost Rate" will show 0.00 for existing data. I did **not** attempt to backfill or fix this — it's a pre-existing gap in the approval flow, out of scope for this issue, and touching business logic there risked exactly what the issue asked to avoid (no changes to the existing tested page/flow). Flagging it here for a possible separate issue.

## Summary
- Adds a new **Stock Take Variance Detailed Report** page (`pharmacy_stock_take_variance_detailed.xhtml`), linked from the existing Stock Take Variance Report via a new "Variance Detailed Report" button.
- Shows per-batch Qty Before/After and Value Before/After at Cost/Retail/Purchase Rate, with footer totals for every Qty/Value column.
- The existing Variance Report page's table, Print, and Download behavior are unchanged — only the new button was added to its header.

## Test plan
- [x] `mvn clean package -DskipTests` — build succeeds, all 207 unit tests pass
- [x] Local redeploy to Payara, verified via Playwright: navigated Stock Take → Variance → Variance Detailed Report for a real stock take (Main Pharmacy, MP//1, 3,110 adjustment lines / 125 report pages)
- [x] Cross-checked footer totals (Qty Before/After, Value Before/After × 3 rates) directly against the database — exact match
- [x] Confirmed the existing Variance Report page still renders correctly and unaffected for the same bill (own footer totals verified separately)
- [x] Checked server log — no errors during testing

Evidence: https://github.com/hmislk/hmis/issues/23139#issuecomment-5345850545

## Summary by CodeRabbit

* **New Features**
  * Added a detailed pharmacy stock-take variance report for selected snapshots.
  * View item, batch, quantity, rate, and cost, retail, and purchase valuation variances.
  * Added quantity and valuation totals with filtering and pagination.
  * Added Excel export, printing, and navigation back to the variance summary.
  * Added guidance when no report is selected.

* **Improvements**
  * Improved report downloads and navigation for a smoother experience.

### DIFF
--- a/src/main/java/com/divudi/bean/pharmacy/PharmacyStockTakeController.java
+++ b/src/main/java/com/divudi/bean/pharmacy/PharmacyStockTakeController.java
@@ -2554,15 +2554,61 @@ public class PharmacyStockTakeController implements Serializable {
         varianceRows.removeIf(vr -> vr.getLastPhysicalQty() == null);
     }
 
-    // Build detailed before/after-adjustment rows for the selected snapshot, sourced from the
-    // posted adjustment bill (forwardReferenceBill of the approved physical count bill) — the
-    // authoritative before/after stock values, not the snapshot/physical-count variance sum.
+    // Build detailed before/after-adjustment rows for the selected snapshot. Same scope as
+    // prepareVarianceRows() — every uploaded snapshot item, not just adjusted ones — but with
+    // before/after quantities and values instead of a variance sum. Qty Before/After come from
+    // the posted adjustment bill's authoritative beforeAdjustmentValue/afterAdjustmentValue where
+    // an adjustment was posted (non-zero variance); for uploaded items with zero variance (no
+    // adjustment line exists), before = after = the uploaded physical count quantity.
     private void prepareVarianceDetailedRows() {
         varianceDetailedRows = new java.util.ArrayList<>();
         if (snapshotBill == null || snapshotBill.getId() == null) {
             return;
         }
 
+        // --- Step 1: load snapshot bill items as scalars (before-state + rates) ---
+        String jpqlSnap = "SELECT bi.id, bi.qty, "
+                + "ib.purcahseRate, ib.retailsaleRate, ib.costRate, "
+                + "ib.batchNo, it.code, it.name, cat.name, df.name "
+                + "FROM BillItem bi "
+                + "LEFT JOIN bi.pharmaceuticalBillItem pbi "
+                + "LEFT JOIN pbi.itemBatch ib "
+                + "LEFT JOIN ib.item it "
+                + "LEFT JOIN it.category cat "
+                + "LEFT JOIN it.dosageForm df "
+                + "WHERE bi.bill.id = :billId "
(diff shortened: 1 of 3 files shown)


buddhika75:
Additional fix pushed: the row height on both the existing Variance Report and the new Variance Detailed Report was unexpectedly large (rows 300-650px tall). Root cause: the "Code" and "Item" `p:column`s used `min-width` instead of `width`; under `.ui-datatable table`'s `table-layout:fixed` rule, `min-width`-only columns collapse to 0 width, forcing text to wrap character-by-character. Fixed by switching all column sizing to the codebase's existing `width="Xem"` attribute convention (matching `pharmacy_income_report.xhtml`), replacing inline `style="width:...px"`.

Also added Qty Variance and Value Variance (Cost/Retail/Purchase Rate) columns to the new detailed report with footer totals, per follow-up request.

Verified in-browser: row heights back to ~57px, all new columns compute correctly and match DB-verified totals.

buddhika75:
Three more fixes pushed, found while investigating a report from the user during review ("old report shows 121 pages for a bill, detailed report shows 44"):

1. **Scope mismatch** — turned out not to be stale data, but a real scope bug: the detailed report only listed lines from the posted adjustment bill (non-zero-variance items only), while the existing Variance Report lists every uploaded snapshot item including zero-variance ones. Reworked the query to match the existing report's scope exactly (snapshot items → uploaded physical-count quantities → overlay adjustment-bill authoritative before/after where posted). Verified both reports now show 121/121 pages for the same bill, cross-checked against the DB (3,002 uploaded items ÷ 25/page).
2. **Rate sourcing** (per CodeRabbit's finding, confirmed via DB) — switched Purchase/Retail/Cost Rate from the adjustment bill's own `PharmaceuticalBillItem` (whose `costRate` is never populated during approval) to `ItemBatch`, which has real data. Cost Rate column now shows correct non-zero values.
3. Added a **Back** button to the detailed report, returning to the Variance Report for the same bill. No Process button — the report has no filter form to submit.

All verified in-browser and against the database.

PR #23195, opened 2026-08-21
fix(pharmacy): correct GRN Summary Report stock amount, filters, and PDF columns

### LINKED ISSUE #23170: GRN Summary Report
- [ ] Print is missing 
- [ ] In GRN Stock Amount isn't displayed
- [ ] After downloded PDF unwanted filters displayed( Department Type,Category,Dosage Form,Supplier)
- [ ] Downloded PDF table column different with report.
- [ ] Downloaded excel unwanted filters displayed(Department Type,Category,Dosage Form)

File Path:/reports/inventoryReports/grn_summary_report.xhtml

buddhika75:
## Summary

Fixes all 5 checklist items in #23170:
- [x] Print is missing
- [x] GRN Stock Amount isn't displayed
- [x] Downloaded PDF showed unwanted filters (Department Type, Category, Dosage Form, Supplier)
- [x] Downloaded PDF table columns didn't match the on-screen report
- [x] Downloaded Excel showed unwanted filters (Department Type, Category, Dosage Form)

This was run via `dev-issue-unattended`, so these calls were made without a live human review beforehand:

1. **Stock Amount computed on the fly instead of fixing upstream population.** `bill.billFinanceDetails.totalPurchaseValue` is only populated by the separate GRN Costing workflow (`GrnCostingController`) or the `PharmacyGrnBifdBackfillService` backfill — not at normal GRN completion. Confirmed via DB query that every recent completed GRN has a `BillFinanceDetails` row but `totalPurchaseValue = NULL`. Rather than changing GRN completion logic across multiple controllers (a much larger, riskier change) or running a data backfill against existing bills (explicitly out of scope for unattended runs), the report now computes "Stock Amount" directly from `BillItemFinanceDetails.valueAtPurchaseRate` — already populated per line item at GRN receipt time for every bill in the DB I checked — batched into one grouped JPQL query per report run.
2. **Left `exportGRNAndDirectPurchaseSummaryReportToPDF()` and `getFiltersForGRNDetailReport()` untouched**, adding new dedicated `exportGRNSummaryReportToPDF()` / `getFiltersForGRNSummaryReport()` methods instead. The old methods are still used by `grn.xhtml`'s own "summary" `reportType` view, which has a different (narrower) column set and a real "Department Type/Category/Dosage Form/Supplier" filter UI of its own — changing them in place would have broken that page.
3. **Print button reuses the existing `grn_summary_view.xhtml` print page** (already built, unpaginated, used by `grn.xhtml`'s own "To Print" button) rather than adding a new print view or wiring `p:printer` directly to the paginated on-screen table (which would only print the current page of results). Its "Back to Report" button still points to `grn.xhtml`, not `grn_summary_report.xhtml` — `PharmacyController` is `@SessionScoped`, so filters/results survive either navigation path; only the return destination differs.

## Root cause detail

- The old `bill.billFinanceDetails != null ? ... : 0` EL check never actually detected a missing aggregate, because `Bill.getBillFinanceDetails()` auto-vivifies a new (empty) `BillFinanceDetails` on first access — the check was always true, but `totalPurchaseValue` on that empty object defaulted to `null`, which `f:convertNumber` renders as blank. Root cause was upstream: the aggregate is simply never populated by the normal GRN flow.
- `getFiltersForGRNDetailReport()` (shared by 13+ call sites) is the correct, full filter set for the older `grn.xhtml` report, which really does have Department Type/Category/Dosage Form/Supplier inputs. `grn_summary_report.xhtml` is a newer, simpler standalone page that was wired to reuse it and the old PDF export by mistake.

## Testing

Verified end-to-end against local Payara + MySQL (`coop` DB), Inward department, GRN Summary Report page, date range covering 44 real GRN/Direct Purchase bills:
- Stock Amount column shows real computed values matching `BillItemFinanceDetails` data (spot-checked one bill against the DB directly).
- PDF export: columns now match the on-screen table exactly; filters section shows only From/To Date, Institution, Site, Department/Store, Payment Mode, Purchase Type.
- Excel export: filters section shows the same scoped set (verified by inspecting the generated `.xlsx` shared-strings XML directly — no "Department Type"/"Category"/"Dosage Form"/"Supplier" present).
- Print button navigates to the existing print page showing all 44 matching rows (unpaginated), with working Print/Back buttons.
- No new errors in `server.log` during the test pass.

Screenshots and full verification notes posted on #23170.

Closes #23170

## Summary by CodeRabbit

* **New Features**
  * Added Print access for the GRN Summary report.
  * Added a dedicated PDF export matching the summary view, including filters, purchase type, payment mode, status, discounts, and stock amounts.
  * Added stock amount details for each bill in the GRN Summary.
  * Enhanced Excel exports to preserve selected filters and support dedicated GRN Summary output.

### DIFF
--- a/src/main/java/com/divudi/bean/pharmacy/PharmacyController.java
+++ b/src/main/java/com/divudi/bean/pharmacy/PharmacyController.java
@@ -210,6 +210,8 @@ public class PharmacyController implements Serializable {
     private List<BillItem> directPurchase;
     private List<PharmacyItemPurchaseDTO> directPurchaseDtos;
     private List<Bill> bills;
+    // GRN Summary Report "Stock Amount" column — bill.id -> summed BillItemFinanceDetails.valueAtPurchaseRate. See #23170.
+    private Map<Long, BigDecimal> grnStockAmountsByBillId;
     List<ItemTransactionSummeryRow> itemTransactionSummeryRows;
     private int managePharamcyReportIndex = -1;
     double persentage;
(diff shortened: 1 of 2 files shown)


buddhika75:
@coderabbitai Fixed in 003d538976 — `postProcessGRNAndDirectPurchaseReportExcel()` now computes both the row shift and the 'Generated On' row placement from the actual filter count (`addMetaDataToExcelSheet()` packs 3 label/value pairs per row) instead of the hardcoded `rowIndex = 5` / `shiftRows(...,7)` tuned only for the Summary Report's 7 filters. Verified by downloading the Summary Report's Excel export and inspecting the raw sheet XML directly: all 7 filter labels and 'Generated On:' now land on distinct, non-overlapping rows, with the header/data rows intact after them.

buddhika75:
@coderabbitai review

Issue #23209, opened 2026-08-21
Stock Report by Item – Align Quantity, Purchase Value and Retail Sale Value Columns


Irani96:
File Path: /pharmacy/pharmacy_report_department_stock_by_item_DTO.xhtml

**Description:**
In the **Stock Report by Item**, the table headers and corresponding total values for **Quantity**, **Purchase Value**, and **Retail Sale Value** are not properly aligned.

**Requirement:**

* Align the **Quantity** column header with its corresponding total/value.
* Align the **Purchase Value** header with the **Purchase Value Total**.
* Align the **Retail Sale Value** header with the **Retail Sale Value Total**.
* Ensure the headers, data values, and total row use consistent column alignment.

**Expected Result:**
The table should have proper vertical alignment between each header, its data, and the corresponding total value.

**Priority:** Medium

Issue #23052, opened 2026-08-16
Batch Stock / Item Stock reports wrap item names instead of scrolling horizontally


buddhika75:
## Summary
On the Batch Stock Report and Stock Report by Item pages (Pharmacy Analytics → Stock Reports), long item names wrap across 2-3 lines because the tables shrink their columns to fit the viewport instead of scrolling horizontally. This inflates row height and makes the report much taller than necessary, especially on the Batch Stock Report which can run to 100+ pages of results.

## Environment
- Branch/commit: `development` @ `6706e89f`
- Pages:
  - Pharmacy Analytics → Stock Reports → Batch Stock — `pharmacy/pharmacy_report_department_stock_by_batch_dto.xhtml`
  - Pharmacy Analytics → Stock Reports → Item Stock — `pharmacy/pharmacy_report_department_stock_by_item_DTO.xhtml`
- Department: Main Pharmacy

## Steps to reproduce
1. Navigate to Pharmacy Analytics → Stock Reports → Batch Stock (or Item Stock).
2. Run the report with default filters.

## Expected
The table should scroll horizontally when its content doesn't fit the viewport, keeping each row single-line and column widths natural to their content.

## Actual
Item/Dosage Form/Type column text wraps to 2-3 lines per row (e.g. "10/0 Ethilon U-7003" wraps across 3 lines; "DIAMOX 250 mg (acetazolamide)" wraps across 2 lines) instead of the table scrolling horizontally.

## Root cause
Neither `p:dataTable` sets `scrollable="true"`:
- `pharmacy/pharmacy_report_department_stock_by_batch_dto.xhtml:132-138` — the table's wrapping `h:panelGroup id="gpBillPreview"` additionally forces `style="min-width: 100%!important;"` (line 130), which actively works against horizontal scrolling by insisting the table span the full container width.
- `pharmacy/pharmacy_report_department_stock_by_item_DTO.xhtml:124` — same missing `scrollable` attribute.

Without `scrollable="true"`, the browser auto-shrinks column widths to fit the viewport rather than allowing overflow, forcing long cell content to wrap.

## Suggested fix
Add `scrollable="true"` (with an appropriate `scrollWidth`) to both `p:dataTable` components, and remove/relax the `min-width: 100%!important` constraint on the Batch Stock Report's wrapper so the table can render at its natural (wider) width with a horizontal scrollbar.

## Evidence
[...]

PR #23229, opened 2026-08-22
23209 stock report by item align quantity purchase value and retail sale value columns

thisarasamuditha:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stock report table alignment when printing and scrolling.
  * Column widths now remain consistent across headers, item rows, and totals.
  * Print layouts are adjusted reliably before the report opens.
  * Improved alignment after sorting, filtering, paging, and other table updates.
  * Updated the totals row to align correctly with corresponding table columns.

### DIFF
--- a/src/main/webapp/pharmacy/pharmacy_report_department_stock_by_item_DTO.xhtml
+++ b/src/main/webapp/pharmacy/pharmacy_report_department_stock_by_item_DTO.xhtml
@@ -140,10 +140,21 @@
                                  content. -->
                             <style>
                                 [id="frm:tbl"] td, [id="frm:tbl"] th { white-space: nowrap; }
+                                /* All three scrollable tables (header / body / footer) must auto-size
+                                   to content AND use border-box so a width set from outerWidth()
+                                   renders as the same box on all three. Leaving the footer out (or
+                                   leaving it content-box) is what let the totals row drift out of
+                                   alignment for long item names - issue #23209. */
                                 [id="frm:tbl"] .ui-datatable-scrollable-header table,
-                                [id="frm:tbl"] .ui-datatable-scrollable-body table {
+                                [id="frm:tbl"] .ui-datatable-scrollable-body table,
+                                [id="frm:tbl"] .ui-datatable-scrollable-footer table {
                                     table-layout: auto !important;
                                     width: auto !important;
+                                }
+                                [id="frm:tbl"] .ui-datatable-scrollable-header th,
+                                [id="frm:tbl"] .ui-datatable-scrollable-body td,
+                                [id="frm:tbl"] .ui-datatable-scrollable-footer td {
+                                    box-sizing: border-box;
                                 }
                             </style>
                             <script>
@@ -158,9 +169,17 @@
                                     if (!$headerCells.length || !$bodyRows.length) {
                                         return;
                                     }
+
+                                    // Clear any previously pinned widths so every cell reports its
+                                    // natural content width again.
                                     $headerCells.css('width', '');
                                     $bodyRows.find('> td').css('width', '');
                                     $footerCells.css('width', '');
+
+                                    // PASS 1 - measure only. Pinning a column before the later columns
+                                    // are measured redistributes space in a table-layout:auto table and
+                                    // corrupts their outerWidth(), so all reads must happen first.
+                                    var widths = [];
                                     $headerCells.each(function(colIndex) {
                                         var maxWidth = $(this).outerWidth() || 0;
                                         $bodyRows.each(function() {
@@ -173,16 +192,38 @@
                                         if (footerWidth > maxWidth) {
                                             maxWidth = footerWidth;
                                         }
-                                        $(this).css('width', maxWidth + 'px');
+                                        widths[colIndex] = maxWidth;
+                                    });
+
+                                    // PASS 2 - apply. With box-sizing:border-box (set in CSS above) the
+                                    // border-box width from outerWidth() maps 1:1 onto css('width').
+                                    $headerCells.each(function(colIndex) {
+                                        var w = widths[colIndex] + 'px';
+                                        $(this).css('width', w);
                                         $bodyRows.each(function() {
-                                            $(this).find('> td').eq(colIndex).css('width', maxWidth + 'px');
+                                            $(this).find('> td').eq(colIndex).css('width', w);
                                         });
-                                        $footerCells.eq(colIndex).css('width', maxWidth + 'px');
+                                        $footerCells.eq(colIndex).css('width', w);
                                     });
                                 }
-                                $(document).ready(function() {
+                                // Run once the DOM is ready, then again after the browser has
+                                // finished layout/paint and PrimeFaces has run its own scrollable
+                                // column sync - on a full-page load (the non-AJAX "Process" submit)
+                                // there is no oncomplete hook, and a bare ready() call measures the
+                                // table before PF has laid out the three scrollable sub-tables.
+                                function alignStockByItemColumnsDeferred() {
                                     alignStockByItemColumns();
-                                });
+                                    if (window.requestAnimationFrame) {
+                                        requestAnimationFrame(function() {
+                                            requestAnimationFrame(alignStockByItemColumns);
+                                        });
+                                    } else {
+                                        setTimeout(alignStockByItemColumns, 50);
+                                    }
+                                    setTimeout(alignStockByItemColumns, 250);
+                                }
+                                $(document).ready(alignStockByItemColumnsDeferred);
+                                $(window).on('load', alignStockByItemColumnsDeferred);
                             </script>
 
                             <p:dataTable
--- a/src/main/webapp/pharmacy/pharmacy_report_department_stock_by_item_DTO.xhtml
+++ b/src/main/webapp/pharmacy/pharmacy_report_department_stock_by_item_DTO.xhtml
@@ -225,11 +225,14 @@
                                 </p:column>
                                 <p:columnGroup type="footer">
                                     <p:row>
-                                        <p:column colspan="4" footerText="Total">
+                                        <p:column/>
+                                        <p:column/>
+                                        <p:column footerText="Total">
                                             <f:facet name="footer">
                                                 <h:outputLabel value="Total" />
                                             </f:facet>
                                         </p:column>
+                                        <p:column/>
                                         <p:column style="text-align: right;" footerText="#{reportsStock.stockPurchaseValue}">
                                             <f:facet name="footer" >
                                                 <h:outputLabel value="#{reportsStock.stockPurchaseValue}" >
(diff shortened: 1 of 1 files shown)


src/main/webapp/pharmacy/pharmacy_report_department_stock_by_item_DTO.xhtml
```diff
@@ -225,11 +277,14 @@
                                 </p:column>
                                 <p:columnGroup type="footer">
                                     <p:row>
-                                        <p:column colspan="4" footerText="Total">
+                                        <p:column/>
+                                        <p:column/>
+                                        <p:column footerText="Total">
                                             <f:facet name="footer">
                                                 <h:outputLabel value="Total" />
                                             </f:facet>
```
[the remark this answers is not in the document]
thisarasamuditha: Ignore it

src/main/webapp/pharmacy/pharmacy_report_department_stock_by_item_DTO.xhtml
```diff
@@ -135,10 +146,48 @@
...
+                                    var $headerCells = $headerTable.find('> thead > tr > th');
+                                    var $bodyRows = $bodyTable.find('> tbody > tr');
+                                    var $footerCells = $footerTable.find('> tfoot > tr > td');
+                                    if (!$headerCells.length || !$bodyRows.length) {
+                                        return;
+                                    }
+                                    $headerCells.css('width', '');
+                                    $bodyRows.find('> td').css('width', '');
+                                    $footerCells.css('width', '');
+                                    $headerCells.each(function(colIndex) {
```
buddhika75, line 183: Measure-then-set is interleaved across columns. The \`.each(colIndex)\` loop pins column N's width on the header/body/footer cells *before* it measures column N+1. Because the tables are \`table-layout: auto\`, fixing an earlier column's width redistributes remaining space and can shrink the *measured* \`outerWidth()\` of later columns below their true content width. Those later columns then get pinned too narrow, and since every cell is \`white-space: nowrap\`, the content overflows/clips (e.g. a large formatted Purchase Value / Retail Sale Value). Fix: do two passes — measure every column into an array first, then apply all widths.

src/main/webapp/pharmacy/pharmacy_report_department_stock_by_item_DTO.xhtml
```diff
@@ -126,7 +126,18 @@
...
+                                 header/footer tables have short content in that column and stay narrow,
+                                 so every column after Item drifts out of alignment (issue #23209).
+                                 alignStockByItemColumns() re-measures the widest cell per column across
+                                 all three tables and pins that width on all three, so they always match
+                                 regardless of item-name length. It reruns after every AJAX update
+                                 (paginate/sort/filter/print) since those redraw the body with different
+                                 content. -->
                             <style>
                                 [id="frm:tbl"] td, [id="frm:tbl"] th { white-space: nowrap; }
                                 [id="frm:tbl"] .ui-datatable-scrollable-header table,
```
buddhika75, line 148: The footer table is missing the CSS override that the header and body tables get. The \`<style>\` block only sets \`table-layout: auto !important; width: auto !important\` on \`.ui-datatable-scrollable-header table\` and \`.ui-datatable-scrollable-body table\` — not \`.ui-datatable-scrollable-footer table\`. In PrimeFaces 14 that footer table stays \`table-layout: fixed\`, so pinning per-\`<td>\` widths on it does not reliably resize it to match the body, and the totals row (the entire point of #23209) can still drift out of alignment.

src/main/webapp/pharmacy/pharmacy_report_department_stock_by_item_DTO.xhtml
```diff
@@ -135,10 +146,48 @@
...
+                                    var $bodyRows = $bodyTable.find('> tbody > tr');
+                                    var $footerCells = $footerTable.find('> tfoot > tr > td');
+                                    if (!$headerCells.length || !$bodyRows.length) {
+                                        return;
+                                    }
+                                    $headerCells.css('width', '');
+                                    $bodyRows.find('> td').css('width', '');
+                                    $footerCells.css('width', '');
+                                    $headerCells.each(function(colIndex) {
+                                        var maxWidth = $(this).outerWidth() || 0;
```
buddhika75, line 184: `outerWidth()` returns the border-box width (content + padding + border), but `.css('width', maxWidth + 'px')` sets the content-box width under the default `box-sizing`. Header `<th>`, body `<td>`, and footer `<td>` typically carry different padding/border, so pinning the same numeric `width` on all three yields three different rendered box widths and the columns stay slightly misaligned. Use `.width()` consistently for both the read and the write, or set `box-sizing: border-box` on these cells.

src/main/webapp/pharmacy/pharmacy_report_department_stock_by_item_DTO.xhtml
```diff
@@ -109,10 +109,10 @@
                                     </p:commandButton>
                                     <p:commandButton styleClass="ui-button-info mx-1" icon="fas fa-print" value="Print"
                                                      actionListener="#{reportsStock.prepareForPrint()}"
-                                                     oncomplete="$('#frm\\:print').click()"
+                                                     oncomplete="alignStockByItemColumns(); $('#frm\\:print').click()"
```
buddhika75, line 112: Print-path realign race. This button realigns after `prepareForPrint()` updates `:frm:tbl`, then clicks the hidden `#print` button. That button's `p:printer` fires `window.print()` client-side immediately on click and does not wait for the `prepareForView()` AJAX to finish, so the realign added to the hidden button's `oncomplete` (line 115) runs only after the print dialog is already built. Net effect: the `oncomplete="alignStockByItemColumns()"` on line 115 is effectively dead for the printout, and the printed table can be misaligned because `prepareForView()` re-rendered the body after the last successful realign.

src/main/webapp/pharmacy/pharmacy_report_department_stock_by_item_DTO.xhtml
```diff
@@ -135,10 +146,48 @@
                                     width: auto !important;
                                 }
                             </style>
+                            <script>
```
buddhika75, line 160: This JS competes with PrimeFaces 14's own scrollable column-width sync. PF renders `<colgroup><col>` elements and re-runs its internal `fixColumnWidths()` after every AJAX update; those `<col>` widths can override the `<td>` inline widths this function sets, and PF's handler races the custom `oncomplete`. The whole mechanism is a bandaid over the `width: auto !important` from #23052 that disabled PF's built-in alignment. A deeper fix is explicit `<p:column width="...">` on every column so all three tables size identically with no JS.

src/main/webapp/pharmacy/pharmacy_report_department_stock_by_item_DTO.xhtml
```diff
@@ -135,10 +146,48 @@
...
+                                    var $footerCells = $footerTable.find('> tfoot > tr > td');
+                                    if (!$headerCells.length || !$bodyRows.length) {
+                                        return;
+                                    }
+                                    $headerCells.css('width', '');
+                                    $bodyRows.find('> td').css('width', '');
+                                    $footerCells.css('width', '');
+                                    $headerCells.each(function(colIndex) {
+                                        var maxWidth = $(this).outerWidth() || 0;
+                                        $bodyRows.each(function() {
```
buddhika75, line 185: Layout thrashing on the print path. The loop interleaves `outerWidth()` reads with `.css('width')` writes across columns x rows, forcing a synchronous reflow every iteration. `prepareForPrint()` renders all rows (paginator off), so on a large stock report this is thousands of forced reflows immediately before `window.print()`, which can visibly freeze the browser. Batch all reads, then all writes.

buddhika75:
## 🚦 Merge Gate: BLOCKED — blocking code-review findings

**Phase 1 (high-effort code review):** 6 findings posted inline (discussion_r3910392454, r3910392678, r3910394029, r3910394147, r3910394308, r3910394471).

**Phase 3 verification (local deploy + Playwright, before Phase 1 was classified as blocking):** I loaded the report and measured column edge positions across the three scrollable tables (header / body / footer):

- **Default data (20 rows):** footer totals row drifts progressively rightward from its columns — ~8px at Code, growing to **~28px at the Sale Value column**.
- **Filtered to one long item name ("NEOLIFE NEOPRINE SHOULDER SUPPORT (M)"):** drift becomes severe — the footer "Total" cell overhangs the body Item column by **~96px**, and the footer Purchase/Sale Value cells are **~130px** offset from the body columns they sum.

This **confirms finding #2** (the `<style>` block adds `table-layout:auto` / `width:auto` only to `.ui-datatable-scrollable-header table` and `.ui-datatable-scrollable-body table` — the footer table is left at PrimeFaces' default `table-layout:fixed`, so `alignStockByItemColumns()` pinning per-`<td>` widths on it doesn't reliably resize it). It also shows **finding #1/#3** (header↔body still off by 7–14px from the interleaved measure-then-set loop and the `outerWidth()`→`width` box-model mismatch).

Net: the PR's stated goal (issue #23209 — align the Quantity / Purchase Value / Retail Sale Value columns, especially the totals row) is **not achieved for the long-item-name case that triggers the bug**. Findings #4–#6 (print-path realign race, PF14 `fixColumnWidths` race, per-iteration reflow on the all-rows print path) were not separately exercised but are consistent with what was observed.

Recommend: give the footer table the same `table-layout:auto` override, and split `alignStockByItemColumns()` into a measure-all pass then an apply-all pass using a consistent box model. Merge-gate stopped here; Phase 2/3 baselines (build, COGS, Cashier Details) did not run.

buddhika75:
## 🚦 Merge Gate: fixes applied (commit a62d971af1)

Addressed the blocking review findings from the previous run:

1. **Footer CSS override** — the footer `<table>` now gets the same `table-layout:auto; width:auto` as the header/body tables.
2. **Box model** — `box-sizing:border-box` on every header/body/footer cell, so the `outerWidth()` value written back via `.css('width')` renders as the same box on all three.
3. **Two-pass alignment** — `alignStockByItemColumns()` now measures every column first, then applies, so pinning an early column no longer corrupts later columns' measurements in a `table-layout:auto` table (this also removes the per-iteration reflow on the print path).
4. **Full-page-load timing** — realign now also runs on `window.load` (rAF + 250ms fallback), since the non-AJAX **Process** submit has no `oncomplete` and a bare `ready()` measured the table before PrimeFaces laid out the scrollable sub-tables.

**Verified live** (Main Pharmacy, Galle coop DB) — header / body / footer column left edges match to **0px** in all three scenarios:
- initial Process load (20 rows)
- filtered to `NEOLIFE NEOPRINE SHOULDER SUPPORT (M)` — the long-name case that previously drifted ~96–130px
- after sorting by Purchase Value

Re-run merge-gate once CI is green for a fresh PASS/BLOCK verdict.

PR #23259, opened 2026-08-23
feat(inward): filter Add Outside Charges item field to outside-fee-eligible items

### LINKED ISSUE #23250: Add Outside Charges: item dropdown lists all services/items instead of only outside-fee-eligible ones
## Summary
The Item field on the Add Outside Charges page lists every non-retired Service/InwardService item, with no restriction to items that actually have an outside fee. Needed: a configurable filter with two modes — (A, default) auto-filter to items whose only site-level fee is the Outside Fee (`FeeType.OtherInstitution`); (B, opt-in) require the item to also be explicitly mapped to the logged department's **site** via a new outside-charge-specific mapping page.

## Environment
- Branch: `development`, commit `ac0afed5`
- Institution: Galle Co Operative Hospital, Inward module
- Local dev instance

## Steps to Reproduce (current behavior)
1. Inward → Search → Admissions → select an admission → Inpatient Dashboard → **Add Outside Charges**.
2. In the Item field, type a search term.
3. Observe: items with no outside fee configured at all still appear in the results.

## Expected vs Actual
- **Expected:** only outside-fee-eligible items appear (per the active mode below).
- **Actual:** every matching Service/InwardService item appears regardless of fee setup.

## Requested Behavior — Two Configurable Modes
- **Mode A (default):** item is listed only if, among its non-retired site-level `ItemFee`s for the current site, the *only* fee type present is `OtherInstitution` (Outside Fee).
- **Mode B (opt-in via new config key):** item must *also* be explicitly mapped on a new outside-charge-specific mapping page, keyed to the department's **site** (`Department.getSite()`) — not the bare department.

## Root Cause
`InwardAdditionalChargeController.completeItem()` (`src/main/java/com/divudi/bean/inward/InwardAdditionalChargeController.java:346-355`) is the autocomplete backing the Item field:

```java
public List<Item> completeItem(String qry) {
    java.util.Map<String, Object> params = new java.util.HashMap<>();
    params.put("name", "%" + qry.toUpperCase() + "%");
    return itemFacade.findByJpql(
            "select i from Item i where i.retired=false "
            + "and (type(i) = InwardService or type(i) = Service) "
            + "and upper(i.name) like :name "
            + "order by i.name",
            params);
}
```

It returns any non-retired `Service`/`InwardService` item matching the typed text — no join against `ItemFee`, no fee-type filter, no mapping check.

Supporting facts:
- Outside fee = `FeeType.OtherInstitution` (label "Out Side Fee") — `src/main/java/com/divudi/core/data/FeeType.java`.
- Site-level item fees are entered on `manage_for_site_item_fees.xhtml`, keyed by `Fee.institution` = the site.
- The general item-mappings page (`manage_department_item_mappings.xhtml`) is backed by `ItemMapping` (`item`, `institution`, `department`, no purpose/category discriminator) — it isn't consulted anywhere in the outside-charge flow, and can't currently distinguish "outside charge mapping" from any other mapping use.
- `Department.site` (`Institution site`) — any outside-charge mapping/filtering must match the department's **site**, not the bare department.
- `InwardAdditionalChargeController` already reads config toggles the same way a new one would (`configOptionController.getBooleanValueByKey(...)`, e.g. lines 474-476) — a new boolean config key fits the existing pattern.

## Evidence
None captured for this issue (navigation path only, no screenshots requested).

buddhika75:
## Decisions made without approval (dev-issue-unattended)

- **Reused the existing `ItemMapping` table** for the new Mode B mapping
  (its `institution` field already models a "site") instead of creating a
  new entity/table, by adding one new boolean column
  `outsideChargeMapping` to distinguish this purpose from `ItemMapping`'s
  other, unrelated uses (OPD/Optician/Collecting-Centre item catalogs,
  "Investigations and Services" institution mapping). A brand-new table
  would be schema design beyond this run's DDL-generation-only exception
  (see Hard limits in the `dev-issue-unattended` skill), so I stayed within
  a single new column on an existing table.
- **Left four live-runtime, projection-only `ItemMapping` queries
  unfiltered** (`completeItemByInstitution`, `fillItemByInstitution`,
  `fillItemLightByInstitution`,
  `ItemController.fillInvestigationsAndServicesMapping`) — these back
  OPD/Optician/Collecting-Centre billing autocomplete and the
  `ITEMS_MAPPED_TO_LOGGED_INSTITUTION` inward item-listing strategy, and as
  projections they don't otherwise need the new column at all. Adding a
  `outsideChargeMapping=false` filter there would make them require the new
  column to exist in the DB the instant this deploys, ahead of the admin
  applying the DDL. I *did* add that filter to the admin-only, full-entity
  `SELECT im FROM ItemMapping im ...` queries (the existing Manage
  Department/Institution Item Mappings pages), since those already require
  the column regardless of any filter (a full-entity fetch reads every
  mapped column) — so the filter there is free correctness with no added
  blast radius.
  **Trade-off:** an item mapped only for outside-charge purposes could
  still surface in those four unrelated pickers if the same site/institution
  also uses them for OPD/Optician/Collecting-Centre/investigation listing.
  This is a narrow, opt-in-only (Mode B, default off) exposure — flagging
  it here for a human to weigh rather than fixing it, since a fix would
  require the new column on paths that don't otherwise need it.
- **Site source is `Department.getSite()`**, not the bare department or the
  bill's "outside institution" — matches the existing scoping pattern in
  `BillBhtController` for site-based fees. Departments with no site
  configured (66 of 95 non-retired departments in the local `coop` DB) keep
  the original, unrestricted item list rather than showing nothing.

## Problem

The Item field on Add Outside Charges listed every non-retired
Service/InwardService item, with no restriction to items that actually have
an outside fee configured (`InwardAdditionalChargeController.completeItem()`
was a blanket type + name-match query, no join to `ItemFee`).

## Fix

- **Mode A (default, always on):** an item is listed only if, among its
  non-retired site-level `ItemFee`s for the logged department's site, the
  *only* fee type present is `FeeType.OtherInstitution` (Outside Fee).
- **Mode B (opt-in via new config key `Inward Outside Charge Requires Item
  Mapping`, department-scoped through `ConfigOptionController`):** the item
  must *also* be explicitly mapped to the site on a new admin page, **Manage
  Outside Charge Item Mappings** (Admin > Manage Items/Services > Item
  Mapping).

## Verification

End-to-end via Playwright + DB against the local `coop` database (Galle
site, Inward department):

- Created three `ItemFee` scenarios at the Galle site: an item with only an
  Outside Fee, an item with only an Own-Institution ("Hospital") fee, and an
  item with both.
- Item field search returned **only** the Outside-Fee-only item; the
  Own-Institution-only item and the mixed item were both excluded; a plain
  no-fee item was also excluded (reproducing the original bug's absence).
- Added the eligible item as a real outside charge and confirmed it
  persisted correctly in `BILLITEM`.
- The new Manage Outside Charge Item Mappings page renders and is wired up
  correctly (Mode B); its full CRUD flow needs the pending DDL column
  applied first (see below), which per this run's hard limits stays a human
  step via the admin UI — not something this run applies, even locally.

## ⚠️ DDL required before/at deploy

This PR adds one new column, `ITEMMAPPING.OUTSIDECHARGEMAPPING`
(`tinyint(1) default 0`). `tmp/createDDL.jdbc` has been regenerated (212
`CREATE TABLE` / 5280 `ADD COLUMN` statements, consistent with the
documented baseline). **Apply it via Admin > Manage Metadata > Add Missing
Fields before or immediately at deploy** — until then, any full-entity
`ItemMapping` read will fail, which affects the pre-existing Manage
Department/Institution Item Mappings admin pages as well as the new one
(same as any other new mapped column, per project policy). The four
runtime billing-autocomplete paths listed above are unaffected either way.

## Documentation

Updated the [Add Outside Charges](https://github.com/hmislk/hmis/wiki/Add-Outside-Charges)
wiki page with a new section describing the filtering behavior, plus a
screenshot showing the Item field returning only the eligible match out of
three same-named items.

Closes #23250

### DIFF
--- a/src/main/java/com/divudi/bean/common/ItemController.java
+++ b/src/main/java/com/divudi/bean/common/ItemController.java
@@ -4266,6 +4266,13 @@ public class ItemController implements Serializable {
                     + " WHERE im.retired = false ";
             HashMap<String, Object> parameters = new HashMap<>();
             if (institution != null) {
+                // NOTE: deliberately not filtering out outsideChargeMapping=true
+                // mappings here (issue #23250) — this is a live runtime item-listing
+                // path (ITEMS_MAPPED_TO_LOGGED_INSTITUTION strategy) and a
+                // projection query that otherwise never needs to touch the new
+                // column; adding the filter would make it require
+                // OUTSIDECHARGEMAPPING to exist in the DB from the moment this
+                // deploys, ahead of the admin's "Add Missing Fields" DDL step.
                 jpql += " and im.institution=:ins ";
                 parameters.put("ins", institution);
             }
--- a/src/main/java/com/divudi/bean/common/ItemMappingController.java
+++ b/src/main/java/com/divudi/bean/common/ItemMappingController.java
@@ -53,6 +53,11 @@ public class ItemMappingController implements Serializable {
     private Item item;
     private ItemMapping selectedItemMapping;
     private List<Item> selectedItems;
+    // Outside-charge-specific mapping page state (issue #23250, Mode B) —
+    // kept separate from `institution`/`department` above so the general
+    // mapping pages and this one don't clobber each other's selection when
+    // both are open in the same session.
+    private Institution outsideChargeSite;
 
     public void addAllSelectedItemsToInstitution() {
         if (selectedItems == null) {
@@ -121,10 +126,14 @@ public class ItemMappingController implements Serializable {
     }
 
     public boolean mappingExists(Item i, Institution ins) {
+        // Excludes outside-charge-only mappings (issue #23250) so this
+        // general-purpose check/reactivate path never hijacks a row created
+        // by the dedicated outside-charge mapping page.
         String jpql = "select pavan "
                 + " from ItemMapping pavan "
                 + " where pavan.institution=:ins "
-                + " and pavan.item=:item ";
+                + " and pavan.item=:item "
+                + " and pavan.outsideChargeMapping=false ";
         Map m = new HashMap<>();
         m.put("ins", ins);
         m.put("item", i);
@@ -139,10 +148,12 @@ public class ItemMappingController implements Serializable {
     }
 
     public ItemMapping findItemMapping(Item i, Institution ins) {
+        // Excludes outside-charge-only mappings (issue #23250) — see mappingExists(Item, Institution).
         String jpql = "select pavan "
                 + " from ItemMapping pavan "
                 + " where pavan.institution=:ins "
-                + " and pavan.item=:item ";
+                + " and pavan.item=:item "
+                + " and pavan.outsideChargeMapping=false ";
         Map m = new HashMap<>();
         m.put("ins", ins);
         m.put("item", i);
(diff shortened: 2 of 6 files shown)


buddhika75:
@coderabbitai review

PR #23261, opened 2026-08-23
fix(inventory-reports): make Add to Stock Bills cost columns visible and total them in exports

### LINKED ISSUE #23224: Add Cost Rate, Cost Value and Total Cost Columns to Add to Stock Bill List Report
**_RH Issue Tracker: 782_**

**_Related issue:  [#17793](https://github.com/hmislk/hmis/issues/17793)_**

---

### **Describe the requirement**

The **Add to Stock Bills Report (under Cost of Goods Sold Report)** should display the **Cost Rate** and **Cost Value** for each item, along with the relevant totals.

Path: `reports/inventoryReports/add_to_stock_bills.xhtml`

### **Expected behavior**

The report should include:

* **Cost Rate** – Cost per unit of the item.
* **Cost Value** – Total cost value for the relevant quantity.
* **Total Cost Value** – Total cost value of all items/records in the report.

The report should allow users to clearly identify and calculate the **cost of goods added to stock**.

### **Screenshots from RH**

Image

buddhika75:
﻿Closes #23224

## What was wrong

The report already declared `CODE`, `BATCH NO`, `QTY`, `COST RATE`, `COST VALUE`, `PURCHASE RATE` and `PURCHASE VALUE` columns — but none of them carried a `width`. PrimeFaces renders the DataTable with `table-layout: fixed`, so the columns that *did* declare a width consumed the whole table and these seven collapsed to **0px**.

Measured on the running app before the fix, at a 1600px viewport:

```
NAME             w=832
CODE             w=0
BATCH NO         w=0
QTY              w=0
COST RATE        w=0
COST VALUE       w=0
PURCHASE RATE    w=0
PURCHASE VALUE   w=0
MRP              w=160
```

The cells rendered their values (`130.00` was in the DOM) — they just had no width, so the user saw the item name run straight into MRP. That is exactly the screenshot on the issue.

## What changed

**1. Column widths (`add_to_stock_bills.xhtml`).** Every column now declares a width, so none can collapse. `NAME` drops from `50em` to `20em` so the row is not needlessly wide; the value columns are right-aligned.

**2. Export totals (`PharmacyReportController`).** The Excel export had **no totals row at all**, and the PDF totalled only Net Total. Both now close with a Total row carrying **Total Cost Value**, **Total Purchase Value**, **Net Total** and **Total MRP Value**, and the PDF repeats them in its Summary block.

The totals come from the controller totals accumulated during processing, not from re-summing the exported rows — so they match the on-screen footer, including the negative reference bills and pre-add-to-stock bills that are netted at total level only.

**3. Export column corrections.** In both exports the columns headed `RATE`/`VALUE` actually held MRP; they are relabelled `PURCHASE RATE`/`PURCHASE VALUE` and repointed at the purchase rate. The `MRP`/`MRP Value` columns — previously left blank in Excel and empty rowspan cells in the PDF — are now populated per item. Null quantities and rates no longer throw while writing a row.

These two export methods are shared with `reports/inventoryReports/opd_sale.xhtml`, which has the identical on-screen column set and footer, so it gets the same corrected exports.

## Verification

Built and deployed locally, ran the report for **16 Apr 2026, all institutions/sites/departments** against the local `coop` database (50 bills).

Reconciled against the database:

| | Cost Value | Purchase Value | Net Total |
|---|---|---|---|
| 49 settle-at-cashier bills (SQL) | 8,765.70 | 8,950.61 | 10,286.18 |
| + 1 pre-add-to-stock bill (SQL) | 569.04 | 675.62 | 575.90 |
| **Expected** | **9,334.74** | **9,626.23** | **10,862.08** |
| Report footer | 9,334.75 | 9,626.23 | 10,862.08 |
| Excel Total row | 9334.7473 | 9626.2274 | 10862.08 |
| PDF Total row + Summary | 9334.75 | 9626.23 | 10862.08 |

Column widths after the fix: `COST RATE` 144px, `COST VALUE` 160px, `PURCHASE RATE` 144px, `PURCHASE VALUE` 160px — previously all 0.

### The report, cost columns now visible

Add to Stock Bills report with Cost Rate and Cost Value

### PDF export Total row and Summary

PDF export totals

Excel `Total` row (row 103) as written:

```
A=Total | H=9334.7473 | J=9626.227400000002 | K=10862.080000000002 | N=11282.080000000002
```

## Documentation

Wiki page updated: [Add-to-Stock-Bills-Configuration](https://github.com/hmislk/hmis/wiki/Add-to-Stock-Bills-Configuration)

It now documents the 15 export columns and the Total row, and carries both screenshots above. Two statements on that page did not match the system and were corrected: cost rate has **no** fallback to purchase rate (a batch with no cost rate contributes 0.00), and exports are **not** limited by the page size.

Also added [§80 to the Playwright E2E workflow doc](https://github.com/hmislk/hmis/blob/23224-add-to-stock-bills-cost-columns/developer_docs/testing/playwright-e2e-workflow.md) — a fixed-layout DataTable column can be present in an accessibility snapshot and still be 0px wide, which is how this issue first read as "already implemented". Measure `offsetWidth` before concluding a column exists.

## Noted, not changed

- `reports/inventoryReports/opd_sale.xhtml` has the **same** zero-width column defect, plus an orphaned `p:columnGroup type="header"` declaring 15 columns for a 17-column table. Out of scope for this issue; worth a follow-up.
- The PDF is still titled "Pharmacy Sales Report" even when launched from Add to Stock Bills (the `selectReportName` actionListener result is unused). Left alone deliberately — it affects both reports sharing the method.

## Summary by CodeRabbit

* **New Features**
  * Pharmacy PDF and Excel reports now show purchase, cost, net, and MRP rates and values in clearly separated columns.
  * Excel reports include an itemized MRP column and a totals row.
  * PDF summaries now include cost, purchase, and MRP totals.

* **Bug Fixes**
  * Reports handle missing payment methods and numeric values more reliably.
  * Inventory report columns now use improved widths, alignment, and readability.

* **Documentation**
  * Added guidance for diagnosing zero-width columns in fixed-layout data tables.

### DIFF
--- a/developer_docs/testing/playwright-e2e-workflow.md
+++ b/developer_docs/testing/playwright-e2e-workflow.md
@@ -2061,6 +2061,28 @@ window.setTimeout = function (fn, delay) { return delay >= 2500 ? 0 : window.__o
 The real growl then stays rendered for the screenshot (nothing about the message is faked — only the dismissal is
 deferred). Reload the page afterwards to drop the patch. Verified while testing issue #23206.
 
+## 80. A `p:dataTable` column present in `browser_snapshot` can still be 0px wide and invisible to the user — measure `offsetWidth`
+
+PrimeFaces renders its DataTable with `table-layout: fixed`. Under that layout the browser honours the columns that
+carry an explicit `width` first and hands the leftovers to the rest — and when the widths already declared exceed the
+table width, "the rest" gets **zero**. Those columns still render their `<th>`/`<td>` with the correct text, so they
+appear in `browser_snapshot`, in `get_page_text`, and in `innerText`. They are simply not on screen.
+
+This is how issue #23224 was misdiagnosed at first: the accessibility snapshot listed all 17 headers including
+`COST RATE` and `COST VALUE`, the footer totals were right, and the DB reconciled — so the report looked finished. The
+columns the user was asking for were 0px wide, exactly as their screenshot showed.
+
+Whenever a report "already has" a column a user says is missing, measure before concluding:
+
+```js
+[...document.querySelectorAll('.ui-datatable thead th')]
+    .map(th => ({ h: th.innerText.trim(), w: Math.round(th.offsetWidth) }))
+```
+
+Any `w: 0` is an invisible column. The fix is in the XHTML, not the test: once *any* `p:column` on a fixed-layout table
+declares a `width`, **every** column must declare one, or the undeclared ones collapse. Widening the viewport does not
+reveal them — it makes it worse, because the extra space goes to the columns that did declare a width.
+
 ## Some PrimeFaces buttons need a jQuery-triggered click
 
 Most `p:commandButton`s submit fine with a normal Playwright click — including
--- a/src/main/java/com/divudi/bean/report/PharmacyReportController.java
+++ b/src/main/java/com/divudi/bean/report/PharmacyReportController.java
@@ -14686,7 +14686,8 @@ public class PharmacyReportController implements Serializable {
                 addSpanningCell(table, String.format("%.2f", billNetTotal), rowSpan, createNetTotalCell(""));
 
                 // This cell acts as a placeholder for the payment methods that will be added below
-                PdfPCell paymentPlaceholder = createDataCell(billDto.getPaymentMethod().toString());
+                PdfPCell paymentPlaceholder = createDataCell(
+                        billDto.getPaymentMethod() != null ? billDto.getPaymentMethod().toString() : "");
                 paymentPlaceholder.setRowspan(rowSpan);
                 table.addCell(paymentPlaceholder);
 
@@ -14760,10 +14761,14 @@ public class PharmacyReportController implements Serializable {
     private void addPaymentBreakdownRows(PdfPTable table, CostOfGoodSoldBillDTO billDto) {
         Map<PaymentMethod, Double> paymentBreakdown = getPaymentBreakdown(billDto);
 
+        // A bill can carry no payment method at all - the export must still
+        // produce its row rather than abort the whole PDF part-written.
+        PaymentMethod billPaymentMethod = billDto.getPaymentMethod();
+
         // Handle single payment method case
-        if (!"MultiplePaymentMethods".equals(billDto.getPaymentMethod().toString())) {
-            String methodName = billDto.getPaymentMethod().toString();
-            double total = billDto.getNetTotal();
+        if (billPaymentMethod != PaymentMethod.MultiplePaymentMethods) {
+            String methodName = billPaymentMethod != null ? billPaymentMethod.toString() : "";
+            double total = billDto.getNetTotal() != null ? billDto.getNetTotal() : 0.0;
             addPaymentMethodRow(table, methodName, total);
         } else {
             for (Map.Entry<PaymentMethod, Double> entry : paymentBreakdown.entrySet()) {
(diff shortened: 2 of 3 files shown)


buddhika75:
@coderabbitai review

buddhika75:
## Self-review (no bot review available)

CodeRabbit has been stuck on "Review in progress" for over an hour with no review submitted, and Codex reported "You have reached your Codex usage limits for code reviews". So this PR got a manual review instead. Two real defects came out of it, both now fixed in 01615ce.

### 1. PDF export aborted part-written on a bill with no payment method

The previous commit guarded `Bill.paymentMethod` on the Excel side but left two unguarded dereferences in the PDF path — `addBillRows` (`:14689`) and `addPaymentBreakdownRows` (`:14764`).

The column is nullable and it is not a theoretical case:

```
paymentMethod distribution, PHARMACY_RETAIL_SALE_PRE_TO_SETTLE_AT_CASHIER:
  Cash 96400 | Card 20453 | Credit 89 | NULL 83
  -> 67 of the NULL bills have items and referenceBill_id IS NULL, so they reach the report
```

The failure mode was worse than a plain crash: the NPE landed inside `exportPharmacySalesToPdf`'s try block *after* the content type, the attachment header and part of the table had been written, and `context.responseComplete()` was never reached. The user got a truncated PDF with JSF markup appended and never saw the error growl.

The `MultiplePaymentMethods` test now compares the enum directly rather than `getPaymentMethod().toString()`. `PaymentMethod` does not override `toString()`, so that is the identical test minus the dereference. `getNetTotal()` on the same line was unboxed unguarded and is defaulted too.

### 2. The totals-row javadoc I added was wrong

It said negative reference bills are "netted at total level only". They are not — `retrieveNegativeReferenceBills` ends with `cogsBillDtos.addAll(negativeBills)`, so they are exported as ordinary rows. Only the bill's `netTotal` is flipped negative; the item quantities and rates stay **positive** while the accumulated totals subtract them. 283 bills in this database would enter via that path.

So `=SUM()` over a value column can legitimately disagree with the Total row (also true of bills with no items, which `populateDataRows` skips but which still count toward the net total). The javadoc now states this rather than implying the columns reconcile.

### Verification

Re-ran against **10 Apr 2026**, a range deliberately chosen because it contains `ETU/SALE/18198` (null payment method) — 56 bills.

| | Cost Value | Purchase Value | Net Total |
|---|---|---|---|
| Report footer | 15,898.91 | 16,006.10 | 19,378.28 |
| Excel Total row (row 138, 137 data rows) | 15898.9146 | 16006.1047 | 19378.28 |
| PDF Total row + Summary | 15898.91 | 16006.10 | 19378.28 |

The PDF now renders all 9 pages with the Summary block intact, and the null-payment bill appears with a blank payment cell:

```
10 April 2026 | ETU/SALE/18198 | Normal Saline (SPC) | 1 | 130.00 | 130.00 | 130.00 | 130.00 | 223.00 |  | 150.00 | 150.00 | 0.00
```

The 16 Apr range used for the original verification contained no negative-reference and no null-payment bills, which is why neither defect surfaced there.

### Reviewed and found correct

- **iText rowspan placement.** `PdfPTable.addCell` calls `skipColsWithRowspanAbove()`, so the per-row MRP cells land in columns 13/14. Confirmed in the rendered output for 1-, 2- and 3-item bills.
- **PDF `grandTotal` equals the accumulated `netTotal`** — both range over the same `cogsBillDtos`.
- **`opd_sale.xhtml` is not regressed** by the shared-exporter change; its table already showed PURCHASE/MRP columns, so the remap fixes it there too.

### Still not changed, deliberately

- `opd_sale.xhtml` has the same zero-width column defect plus an orphaned `p:columnGroup` declaring 15 columns for a 17-column table.
- Exports are internally titled "Pharmacy Sales Report" regardless of which report launched them.
- Fractional quantities truncate in the PDF (`QTY x RATE != VALUE` on that row) where the screen rounds. Only 2 such items exist in this database.

Issue #23165, opened 2026-08-20
Barcode-friendly admission (BHT) search with auto-select across Inward/Pharmacy admission-search pages


buddhika75:
## Summary

Across the Inward/Inpatient menu and the Pharmacy → Inpatient Medication Management menu, ~20 pages ask the user to search for and select an admission (BHT) before doing something with it (edit details, issue medicine, add services, room change, billing, reports, etc.). These pages print the patient's PHN/MRN barcode on multiple documents (wristband, bill, etc.), and staff commonly use a barcode reader to scan it into the search box. Today that experience is inconsistent and mostly not barcode-optimized:

- Only one page (`pharmacy_bill_issue_bht.xhtml`) has any Enter-key handling for a scanner's trailing Enter, and even that only avoids a dead keystroke — it doesn't auto-select.
- Every other page requires the user to manually pick the row from the autocomplete dropdown even when a scan can only ever match one admission.
- The "Patient Selection" panel markup (autocomplete + PHN/BHT/Name/Room/Status columns) is hand-copied per page rather than shared, which is how the inconsistency happened and will keep happening as pages are added.
- MRN (`patient.code`) is searchable on some pages but silently missing from others, so a scanned MRN barcode works on one page and does nothing on the next.

## Goal

Make every admission-search page behave the same way for scanner input:

1. **Support PHN, BHT number, and MRN** as exact-match identifiers everywhere an admission search exists (MRN currently missing from several pages — see below).
2. **Auto-select on scan**: if the query is an exact PHN/BHT/MRN match and returns exactly **one** admission, select it automatically and immediately trigger the page's existing next-step action (whatever "Continue"/"Select" already does on that page) — no extra click needed.
3. **Manual pick on ambiguity**: if an exact-identifier query returns more than one match (e.g. a patient with multiple historical admissions on a page that doesn't filter to active-only), show the normal autocomplete dropdown for the user to pick.
4. **No change for free-text search**: partial name/phone queries always show the dropdown for manual selection, even if only one row currently matches — never auto-advance while the user is mid-typing a name.

## Explicitly out of scope

- **Do not change which admissions each page's query returns.** Some pages intentionally search active-only admissions, some search only not-yet-payment-finalized admissions, some search everything (billing/report pages). That per-page filtering logic stays exactly as-is. The only query change is additive: add MRN (`patient.code`) as a searchable field wherever it's missing, without touching the existing discharged/finalized/active filters.
[...]

PR #23433, opened 2026-09-02
feat(inward,pharmacy): wire shared admission search into billing pages

buddhika75:
## Summary

Continues #23165's barcode-friendly admission search rollout, following PR #23426 (room change pages).

Two commits:
1. **Composite enhancement** — adds an optional `itemSelectListener` attribute to the shared `admcc:admission_search` composite, defaulting to a safe no-op. Needed because two of the billing pages below attach real business logic (not just the model value) directly to the autocomplete's own selection event.
2. **9 pages migrated**: `pharmacy_bill_issue_bht_old.xhtml` (the QA-only comparison page left over from #23165's own scope) plus 8 of the "Family C" billing pages that have a genuine "search for a BHT" step:
   - `inward_bill_payment.xhtml`, `inward_bill_payment_1.xhtml`, `inward_bill_refund.xhtml`
   - `inward_bill_post_final_payment.xhtml`, `inward_bill_post_final_payment_refund.xhtml`
   - `inward_bill_deposit.xhtml`
   - `inward_bill_intrim_finalized.xhtml`, `inward_bill_intrim_finalized_error_correction.xhtml`

All these pages' search methods (`admissionController.completePatient` / `completePatientAll`) were already MRN-searchable from the original pilot PR — no JPQL changes needed here.

## Decisions made without approval / judgment calls

1. **Scope narrowed from the full "Family C" list.** The issue's own tracking comment lists ~20 billing/report pages as "Family C — lower priority." Investigating them, two shapes emerged:
   - Pages with a genuine "find and select one BHT, then act" step — these 8, migrated here.
   - Pages where the BHT box is one optional filter among several on a report with no single-admission-selection step (e.g. `bht_deposit_by_discharge_date.xhtml` and its `_and_created_date` sibling have **no BHT search field at all** — pure date/credit-company filtered aggregate reports) — these are a different UX shape than what this issue's acceptance criteria describe, and are left for a separate follow-up rather than forced into this pattern.
2. **Composite extended with `itemSelectListener`** rather than leaving `inward_bill_payment_1.xhtml` and `inward_bill_intrim_finalized_error_correction.xhtml` unmigrated. Confirmed via a regression check against the pilot page (`inward_edit_bht.xhtml`) that the new default no-op doesn't affect any of the ~15 already-migrated pages.
3. **`continueClientId` deliberately omitted on `inward_bill_payment_1.xhtml`.** Its "Pay" button submits a real payment; auto-clicking it on a barcode scan would fire a financial transaction without staff review. Used `itemSelectListener` instead so scanning still auto-populates the due amount and patient details, but the actual payment stays a deliberate click. Same reasoning applies to why `inward_bill_intrim_finalized_error_correction.xhtml` uses `itemSelectListener` alone (loading a read-only report, not a mutation) rather than `continueClientId` (it has no separate Continue button to click anyway).
4. **Verified each "select" action is read-only setup before wiring auto-advance** on the other 6 pages (`bhtListener`, `selectBhtListener`, `selectBhtListenerForRefund`, `createTablesFinalized`) — all just load/compute display state, none mutate persisted data.
5. **Fixed a layout regression**, not just the search UX: the composite's `width:100%` collapses to a tiny box inside an `h:panelGrid`'s auto-sized column on 3 of these pages (they don't use a Bootstrap `col-*` div wrapper like the other migrated pages). Added an explicit `min-width: 350px` on each affected page's BHT panel.

## Verified

End-to-end with Playwright against local Payara + MySQL (`coop` dev DB):
- **`inward_bill_payment.xhtml`**: exact BHT scan auto-selects and auto-advances to Payment Details; the due-amount panel correctly reflects `bhtListener()`'s output ("Payment Details Are Not Finalized").
- **`inward_bill_intrim_finalized.xhtml`**: exact BHT scan auto-selects and auto-clicks Display; Room Detail tab shows real charge data (e.g. Room Charge 11,600) for the scanned BHT, not stale/default data.
- **`inward_bill_payment_1.xhtml`**: scanning a BHT updates the due amount and patient-detail panel but does **not** auto-click Pay — confirmed by scanning a second, different BHT and observing the panel update in place with no navigation.
- **`inward_bill_intrim_finalized_error_correction.xhtml`**: scanning a BHT loads its finalized bill tables immediately (matching the page's original itemSelect behavior), search box renders at the corrected width.
- The 5 remaining Bootstrap-layout pages (refund, post-final-payment, post-final-payment-refund, deposit, old pharmacy) visually confirmed rendering correctly at full width with the shared composite.
- No new errors in `server.log` or the browser console across the whole verification pass.

## Follow-up (not in this PR)

Still outstanding from the issue's original checklist:
- The remaining Family C **report** pages (`inward_bill_payment.xhtml`'s siblings' report-only counterparts, `inward_report_*` pages, `bht_deposit_by_*` variants) — need per-page investigation to classify wizard-vs-filter shape, same as this PR did.
- `bht_deposit_by_discharge_date.xhtml` and `bht_deposit_by_discharge_date_and_created_date.xhtml` specifically have no BHT search field at all and don't fit this issue's pattern — flagged for the issue owner to confirm they're genuinely out of scope rather than silently skipped.

Closes nothing on its own — tracked as part of #23165.

### DIFF
--- a/src/main/java/com/divudi/bean/common/AdmissionSearchSupportController.java
+++ b/src/main/java/com/divudi/bean/common/AdmissionSearchSupportController.java
@@ -1,0 +1,40 @@
+/*
+ * Author : Dr. M H B Ariyaratne
+ *
+ * Consultant (Health Informatics)
+ * (94) 71 5812399
+ * Email : buddhika.ari@gmail.com
+ *
+ */
+package com.divudi.bean.common;
+
+import java.io.Serializable;
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Named;
+
+/**
+ * Backing support for the shared {@code ezcomp/inpatient/admission_search.xhtml}
+ * composite, used across Inward/Pharmacy/Theater/Store admission-search pages
+ * (issue #23165).
+ *
+ * The composite's optional {@code itemSelectListener} attribute needs a
+ * default method to bind to when a calling page doesn't provide its own
+ * "on select" business logic - {@link #noop()} is that default. Stateless
+ * and side-effect free by design, so application scope is safe.
+ */
+@Named
+@ApplicationScoped
+public class AdmissionSearchSupportController implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * Default {@code itemSelectListener} for {@code admcc:admission_search}.
+     * Intentionally does nothing - pages that need real logic to run when an
+     * admission is selected pass their own listener instead.
+     */
+    public void noop() {
+        // Intentionally empty.
+    }
+
+}
--- a/src/main/webapp/inward/inward_bill_deposit.xhtml
+++ b/src/main/webapp/inward/inward_bill_deposit.xhtml
@@ -9,14 +9,15 @@
                 xmlns:bill="http://xmlns.jcp.org/jsf/composite/inward/bill/payment"
                 xmlns:prints="http://xmlns.jcp.org/jsf/composite/ezcomp/prints"
                 xmlns:in="http://xmlns.jcp.org/jsf/composite/inward"
-                xmlns:pa="http://xmlns.jcp.org/jsf/composite/paymentMethod">
+                xmlns:pa="http://xmlns.jcp.org/jsf/composite/paymentMethod"
+                xmlns:admcc="http://xmlns.jcp.org/jsf/composite/ezcomp/inpatient">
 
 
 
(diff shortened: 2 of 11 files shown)


buddhika75:
Note: PR #23440 (the remaining Family C report pages) depends on this PR's `itemSelectListener` composite attribute and must be merged **after** this one. Flagging here so merge order is clear from both sides.

Issue #23523, opened 2026-09-06
Settle on the inward service bill page fails silently when the admission has no room


buddhika75:
## What happens

On **Inward → Services & Items → Add Services & Investigations**
(`inward_bill_service.xhtml`), clicking **Settle** for a patient whose
admission has no current room — or whose current room has no room facility
charge — does **nothing at all**. No bill, no growl, no error, nothing in
`server.log`. The user has no way to tell what is wrong.

Every other rejection path on this page explains itself ("Please select
Staff", "Sorry Patient is Discharged!!!", "Please add the Patient DOB", …).
These two do not.

## Where

`BillBhtController.errorCheck()`, lines 1158-1166:

```java
if ((getPatientEncounter().getAdmissionType().isRoomChargesAllowed() && !isBabyAdmission())
        || getPatientEncounter().getCurrentPatientRoom() != null) {

    if (getPatientEncounter().getCurrentPatientRoom() == null) {
        return true;                          // <-- no message
    }

    if (getPatientEncounter().getCurrentPatientRoom().getRoomFacilityCharge() == null) {
        return true;                          // <-- no message
    }
}
```

`settleBill()` calls `errorCheck()` and returns silently on `true`, so both of
these surface to the user as a dead button.

## Why it matters

`settleBill()` immediately afterwards does

```java
settleBill(getPatientEncounter().getCurrentPatientRoom().getRoomFacilityCharge().getDepartment(), ...)
```

so these two guards are load-bearing — they are preventing an NPE, not being
over-cautious. The guards are right; only the silence is wrong.

## Suggested fix

Add a `JsfUtil.addErrorMessage(...)` to each, in the style of the surrounding
checks — e.g. "Cannot settle: this admission has no current room. Assign a room
first." and "Cannot settle: the patient's current room has no room facility
charge configured."

## Notes

Found while re-testing #21931 / #21971. I originally mis-reported this as
"Settle silently does nothing" in #23519 — that report was about a *different*
cause (a "Please select Staff" growl I had missed) and has been retracted
there. This one is real, but I did not hit it with a live patient; it is read
from the code plus the surrounding behaviour, so the repro below needs an
admission actually in that state.

**Repro:** admit a patient under an admission type with `roomChargesAllowed`,
do not assign a room (or assign one whose room facility charge is null), then
add a service and press Settle.

Issue #23570, opened 2026-09-07
Cannot add services to a day-case admission: billing requires a room the admission will never have


buddhika75:
### Summary

Adding services to an admission currently assumes the patient is in a room. That assumption does not hold for **day cases** — the patient is admitted and discharged in the same visit and never occupies a bed — but they still have services to bill.

### Current behaviour

`BillBhtController` gates every billing path on:

```java
(patientEncounter.getAdmissionType().isRoomChargesAllowed() && !isBabyAdmission())
        || patientEncounter.getCurrentPatientRoom() != null
```

This expression is repeated in five places (`settleBill`, `errorCheck`, `addToBill`, the bill-fee construction, and `feeChanged`). It asks *"what kind of admission is this?"* when the question the screen actually needs answered is *"does this encounter have a room we can bill against?"*.

Consequences:

- A day-case admission type with **Room Charges Allowed** ticked demands a room that will never exist. Add Service reports *"Please Set Room or Bed For This Patient"* and Settle reports *"Cannot settle: this admission has no current room. Assign a room first."* (the message added in #23523) — but there is no room to assign, so the admission simply cannot be billed.
- Whether it works at all therefore depends on an admin having left one checkbox unticked on the admission type, not on anything structural.
- #23509 had to bolt a `!isBabyAdmission()` term onto the same expression for exactly the same reason. That is the second special case on one condition; day cases would be the third.

### Proposed change

Replace the repeated expression with named predicates on the encounter:

- `roomExpected(encounter)` — an ordinary inward stay whose admission type takes room charges. False for day cases (`AdmissionTypeEnum.DayCase`), babies, and admission types that do not take room charges.
- `roomAssigned(encounter)` — a room has actually been assigned.
- `roomRequiredForBilling(encounter)` = `roomExpected || roomAssigned`. A room that *has* been assigned must still be fully configured before anything is billed against it, whatever the admission type.
- `feeDepartment(encounter)` — the room's facility-charge department when there is a room, otherwise the encounter's own department.

`AdmissionTypeEnum.DayCase` is already the test `RoomChangeController` uses to exempt day cases from the room-chain rules, so this makes the billing screen agree with room management instead of disagreeing with it.

No behaviour change for ordinary inward stays: a ward patient with no bed assigned is still blocked, which is the intended safety net.

PR #23581, opened 2026-09-07
fix(inward): re-fetch encounter fresh in errorCheck to close stale-room race

### LINKED ISSUE #23568: BillBhtController holds a stale patientEncounter/currentPatientRoom across requests — Settle validation can silently use outdated room data
## Problem

`BillBhtController` is `@SessionScoped` and holds `patientEncounter` as a plain field, set once when a BHT is selected (`navigateToAddServicesFromMenu()` / the patient-search flow) and read via a bare getter (`getPatientEncounter()`, no refresh):

```java
public PatientEncounter getPatientEncounter() {
    return patientEncounter;
}
```

If the encounter's current room changes after that point — e.g. a different staff member (or the same user in a second tab of the same login session) uses **Patient Room Details → Remove Room** or a Room Change action, which goes through `RoomChangeController` and `patientEncounterFacade.edit(encounter)` — `BillBhtController`'s already-held `patientEncounter`/`currentPatientRoom` reference does not observe that change. A later `errorCheck()`/`settleBill()` call in the original tab reads the stale, no-longer-current room.

## Why it matters

This affects `errorCheck()`'s new room-validity guards added in #23523 (PR #23567): if the room becomes genuinely invalid (removed, or its facility charge cleared) after the encounter was loaded but before Settle is clicked, the guard can silently see the *old*, still-valid-looking room and let `settleBill()` proceed — settling the bill against stale room data instead of either blocking with the new error message or using the current room. This is not just a test artifact: any real concurrent-user scenario (two staff members working the same admission, or the same user switching tabs) can trigger it.

## Evidence

Confirmed empirically while verifying #23523:
1. Added a service to a real BHT's cart on `inward_bill_service.xhtml` (room valid at that point).
2. In a second tab of the *same login session*, used **Room Details → Remove Room** (a real app action, not raw SQL) to null the encounter's `currentPatientRoom` — confirmed via DB that `CURRENTPATIENTROOM_ID` was genuinely `NULL` afterward.
3. Back in the first tab, clicked **Settle**. Expected the new "Cannot settle: this admission has no current room" message; instead got `Bill Saved`, and a real `BILL` row was created using the old room.

Flagged independently by CodeRabbit on PR #23567 (reviewing `developer_docs/testing/playwright-e2e-workflow.md` §96, which documented this as a testing-workflow limitation): "this reproduction is not only a Playwright limitation... track it as a separate production correctness issue instead of documenting it as test-only." This issue is that follow-up.

## Suggested fix (not designed here — needs discussion)

Some form of refreshing `patientEncounter` (or at least `currentPatientRoom`) immediately before the settlement validation in `errorCheck()`/`settleBill()`, rather than trusting the session-held reference. Exact approach (targeted `em.refresh`, re-`find()`, or something else) needs discussion — this project's own notes flag that blind EclipseLink cache eviction has proven unreliable elsewhere in this codebase (see the "Native settle poisons the Bill entity" pattern), and `BillBhtController`'s session-scoped field is relied on elsewhere in the same class, so a change here needs care to avoid breaking the normal (non-racing) flow or introducing a performance regression from refreshing on every Settle click.

## Scope note

Out of scope for #23523/PR #23567, which only added the missing error messages for guards that were already correctly blocking (just silently). This issue tracks the deeper staleness problem those guards can still miss.

buddhika75:
## What

`BillBhtController` is `@SessionScoped` and held `patientEncounter` as a plain field, set once
when a BHT is opened and never refreshed. If a different tab/session (a different staff member,
or the same user in a second tab of the same login session) changed the encounter's current room
after that point — e.g. via **Patient Room Details → Remove Room**, which goes through
`RoomChangeController` and `patientEncounterFacade.edit()` on a *different* `PatientEncounter`
instance in a different persistence context — `errorCheck()`'s room-validity guards (added in
#23523 / PR #23567) and `settleBill()`'s second, independent read of `getCurrentPatientRoom()`
would both keep reading the stale, no-longer-current room, letting a bill get saved against
out-of-date room data instead of being blocked.

## Fix

- Injected `PatientEncounterFacade` into `BillBhtController` (same `@EJB` pattern as the existing
  `PatientFacade`).
- At the top of `errorCheck()`, right after the existing null-check, re-fetch a fresh
  `PatientEncounter` via `findWithoutCache(id)` — the same cache-bypassing pattern already used
  two lines below it for `Patient` — and **reassign** the session field (`this.patientEncounter =
  freshPatientEncounter`), rather than just using a local variable.
- Reassigning matters: it's what makes `settleBill()`'s later, independent read of
  `getCurrentPatientRoom()` (used to pick `matrixDepartment`) also see the fresh state within the
  same Settle click — a purely-local fix inside `errorCheck()` wouldn't have closed that second
  read.
- `findWithoutCache` bypasses EclipseLink L1/L2 cache; a targeted `em.refresh()` on the old field
  was **not** used since that field is detached from the current request's EntityManager (it was
  loaded in a prior HTTP request) and `em.refresh()` on a non-managed entity throws
  `IllegalArgumentException`.

Also includes a small clarification to the `dev-issue` skill: step 7's state-changing setup
actions that exercise an *already-approved* fix are a local-testing-environment judgment call
(same category as picking which department/record to use), not a fresh discussion gate — only
step 2a's reproduction of an *unconfirmed* bug requires asking first. This was ambiguous in the
skill text and caused an unnecessary pause during this issue's verification.

## Verification

Reproduced the exact two-tab race described in the issue against the **fixed** build, on local
Payara against the local `coop` DB, using a synthetic test admission (**BHT/57817**, encounter
13861567 — patient "Test Multi Surgery Patient", clearly non-production test data):

1. **Tab 1** — menu path *Inward → Search → Admissions → BHT/57817 → Inpatient Dashboard → Add
   Services & Investigations* (`inward_bill_service.xhtml`): added an X-Ray item to the cart while
   the room was valid.

   Item added while room is valid

2. **Tab 2**, same login session — menu path *Inward → Search → Admissions → BHT/57817 →
   Inpatient Dashboard → Room Details*: used the real **Remove Room** action on the active room.
   Confirmed in the DB that `PATIENTENCOUNTER.CURRENTPATIENTROOM_ID` went `NULL` and the room row
   was retired.

   Room removed from a second session

3. **Back in Tab 1**: clicked **Settle**. With the fix, the guard correctly sees the fresh state
   and blocks:

   > Cannot settle: this admission has no current room. Assign a room first.

   Settle now blocked with fresh state

   Confirmed via DB query (`SELECT ... FROM BILL WHERE PATIENTENCOUNTER_ID = 13861567`) that no
   new `BILL` row was created for the blocked attempt.

Also ran the full Maven build (`mvn clean package`) — 261 unit tests pass, no compile errors.

## Documentation

No wiki page update — this is an internal correctness fix with no new user-visible page or
workflow. The error message itself already existed from #23523/PR #23567; this change only makes
it fire correctly for the stale-encounter race instead of being silently bypassed.

Fixes #23568

### DIFF
--- a/.claude/skills/dev-issue/SKILL.md
+++ b/.claude/skills/dev-issue/SKILL.md
@@ -15,11 +15,24 @@ argument-hint: "<issue-number>"
 
 Invoking this skill is the explicit authorization for every commit/push/PR
 step below — do not re-ask before each one. Discussion gates (steps 2a
-non-repro case and any state-changing reproduction step, 3, 4's environment
-choice only, 14) are the points where you pause for the user. Everything
-else in 2a and 4 (which department/record to use against local test data)
-is a local-testing-environment choice, not a product decision — decide it
-yourself and say what you picked, rather than pausing.
+non-repro case and any state-changing step taken there to prove an
+*unconfirmed* bug, 3, 4's environment choice only, 14) are the points where
+you pause for the user. Everything else in 2a and 4 (which department/record
+to use against local test data) is a local-testing-environment choice, not a
+product decision — decide it yourself and say what you picked, rather than
+pausing.
+
+That 2a gate is narrow and does not extend to step 7. 2a's risk is spending
+effort chasing a bug that might not be real; once step 3 has been through
+Plan Mode and the user has approved a fix, that risk is gone — exercising
+the approved fix in step 7, including any state-changing UI action needed to
+set up the scenario (e.g. removing a room, changing a status, editing a
+record) against local test data, is the same no-need-to-ask
+local-testing-environment judgment call as picking which department/record
+to use. Decide it yourself, do it through the real app UI (never raw SQL for
+setup — see step 7), and report exactly what you did as evidence. Only ask
+first if the action would reach outside local test data (a remote
+environment, or anything step 4's environment-choice gate already covers).
 
 This authorization also covers `superpowers:writing-plans`' Execution
 Handoff question, if that chain gets invoked anywhere in this flow (e.g.
@@ -149,6 +162,15 @@ errors before moving on.
 
 Run the `playwright-e2e` skill workflow:
 - Login, select the department from step 4
+- If verifying the fix requires putting a record into a specific state first
+  (e.g. a race-condition fix needs a room removed, a status changed, a second
+  record created), do that live through the real app UI yourself — this is
+  the same local-testing-environment judgment call as step 4's
+  department/record choice, not a fresh discussion gate. (Unlike step 2a,
+  which gates state-changing actions because it's proving an *unconfirmed*
+  bug, step 7 is exercising a fix the user already approved in step 3.)
+  Report exactly what you did (menu path, record IDs, before/after DB state)
+  as part of the evidence.
 - **Navigate to the page through the menus, never by URL** — HMIS page state is
   set by the `@SessionScoped` navigation method, so a URL-loaded page renders
   against uninitialised state and produces false findings (`playwright-e2e` §2).
--- a/src/main/java/com/divudi/bean/inward/BillBhtController.java
+++ b/src/main/java/com/divudi/bean/inward/BillBhtController.java
@@ -61,6 +61,7 @@ import com.divudi.core.facade.BillItemFacade;
 import com.divudi.core.facade.EncounterComponentFacade;
 import com.divudi.core.facade.FeeFacade;
 import com.divudi.core.facade.ItemFeeFacade;
+import com.divudi.core.facade.PatientEncounterFacade;
 import com.divudi.core.facade.PatientFacade;
 import com.divudi.core.facade.PatientInvestigationFacade;
 import com.divudi.core.facade.PersonFacade;
@@ -157,6 +158,8 @@ public class BillBhtController implements Serializable {
     private PersonFacade personFacade;
     @EJB
     private PatientFacade patientFacade;
+    @EJB
+    private PatientEncounterFacade patientEncounterFacade;
     @EJB
     private BillComponentFacade billComponentFacade;
     @EJB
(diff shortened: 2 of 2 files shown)


buddhika75:
## Rebased onto latest `development` — regression check for day-case admissions

Good catch, flagged by the reporter: this branch was originally cut before **#23573** ("day-case
admissions bill services without a room", closing #23570) merged into `development`. That PR
rewrote `errorCheck()`'s room-validity gate from an inline condition into named predicates
(`roomExpected`/`roomAssigned`/`roomRequiredForBilling`/`feeDepartment`) so that day cases,
package admissions, and babies can bill without a room. Since this branch touches the very same
method, merging it unrebased would have silently reverted that fix.

**Rebased** onto `origin/development` (now includes #23573 and #23574). The rebase applied with
**no conflicts** — this fix's fresh-refetch block sits right after the existing null-check, before
any of the day-case predicate logic, so it composes cleanly: `errorCheck()` now re-fetches and
reassigns the encounter *before* `roomRequiredForBilling()` and friends run, meaning those
predicates also see fresh data rather than the stale session field — a small bonus, not just a
non-conflict.

### Regression check: day-case admission (no room) still bills correctly

Verified live on the rebased build, local `coop` DB, BHT **OPDCARD/44351** (encounter `13861560`,
admission type "OPD Card" / `DayCase`, `CURRENTPATIENTROOM_ID = NULL`, not discharged):

- Added an X-Ray item via *Inward → Search → Admissions → OPDCARD/44351 → Inpatient Dashboard →
  Add Services & Investigations* — **no** "Please Set Room or Bed For This Patient" error.
- Assigned the Tech Fee's required staff, clicked **Settle** — succeeded with **Bill Saved**
  (`Inward//26/000554`).
- Confirmed via DB: `BILL` row `14080229`, `NETTOTAL = 750`, `PATIENTENCOUNTER_ID = 13861560` —
  same encounter that already had a prior bill (`14080112`, 2026-09-02) from #23573's own
  verification, confirming this is a genuinely room-less, day-case-billable admission both before
  and after this PR.

The original stale-room-race fix itself is untouched by the rebase (identical diff, just
repositioned against the new surrounding code) and was already verified end-to-end before the
rebase — see the issue comment and the PR description above.

Full local build re-run after rebase: `mvn clean package` — 261 unit tests pass, no compile
errors. Redeployed and re-verified both scenarios live rather than trusting the rebase blindly.

Issue #22240, opened 2026-07-19
Inpatient Event History timeline page linked from Inpatient Dashboard (bills + audit events)


buddhika75:
Part of #22232 (Inpatient Event History — Patient Story). Depends on the schema/helper sub-issue (`patientEncounterId` on `AuditEvent`); becomes richer as the instrumentation sub-issues land, but can ship as soon as the schema exists.

## Goal

New **Inpatient Event History** page linked from the Inpatient Dashboard (`/inward/admission_profile.xhtml`) showing one chronological timeline — the full "story" of the admission from admission to final bill.

## Data sources (merged, sorted by timestamp)

1. **Bills** — `Bill` where `patientEncounter = :pe` (and `retired=false`), using `createdAt`, `creater`, `billTypeAtomic`, `netTotal`. Covers services, pharmacy BHT issues, professional fees, deposits/payments, refunds, surgeries, provisional/final bills, and their cancellations/returns. No new recording needed.
2. **Audit events** — `AuditEvent` where `patientEncounterId = :peid`, showing `eventTrigger`, user (resolve `webUserId`), `eventDataTime`, and the existing before/after diff (`AuditEvent.getDifference()`).

## UI

- Page e.g. `/inward/admission_event_history.xhtml` + controller (session-scoped, init in navigation method per project JSF rules — no `f:viewAction`)
- Button/link on `admission_profile.xhtml` ("Event History" / "Patient Story")
- Each timeline row: timestamp, event category (Billing / Room / Admission / Payment / Discharge / Other), description, user, department; audit rows expandable to show the before/after diff; bill rows link to bill view/reprint
- Optional filters: category, date range, user
- Follow `ui-guidelines` skill for layout; privilege-controlled access (new or existing inward privilege)

## Timeline row model

Use a small view DTO (e.g. `InpatientTimelineRow(Date when, String category, String description, String user, Long billId, AuditEvent auditEvent)`) built in the controller by merging the two queries — no schema changes needed beyond the parent's `patientEncounterId`.

## Acceptance criteria
- [ ] Link on Inpatient Dashboard opens the Event History for the selected admission
- [ ] Timeline shows bills + audit events merged chronologically with user attribution
- [ ] Audit rows show before/after diffs; bill rows identify bill type and amount
- [ ] Cancelled bills appear as their own events (original bill + cancellation bill both shown)
- [ ] Page access is privilege-controlled

buddhika75:
## Verified locally (Playwright E2E + DB)

[...]

PR #23766, opened 2026-09-12
fix(inward): log Timed Service removal to the event-history timeline

### LINKED ISSUE #23722: The History is not updated after removing the Timed Service.
File Path: http://124.43.4.189:8010/rhLocal/faces/inward/admission_event_history.xhtml

Current Issue:
After removing a Timed Service, the History is not updated to reflect the removal.

Expected Behavior:

The History should be updated immediately after a Timed Service is removed.

**Issue Num:750**

buddhika75:
## Summary

Fixes #23722 — after removing a Timed Service, the Inpatient Event History timeline (`admission_event_history.xhtml`, the "Patient Story" feature from #22240) never reflected the removal.

**Root cause**: the timeline merges two sources: bills (`Bill` rows filtered to `retired=false`) and encounter-linked `AuditEvent` rows. Removing a Timed Service correctly retires its own `Bill` (`InwardTimedItemController#retireTimedServiceBill`) — but that just makes the row *disappear* from the `retired=false` billing query, with nothing left behind to show a service was ever added or removed. Neither removal entry point (`InwardTimedItemController#removePatientItem` for the ward flow, `SurgeryBillController#removeTimeService` for the Interim Bill / surgery flow) ever called `AuditService`, unlike sibling actions in the same classes (e.g. "Validate Surgery" / "Revert Surgery Validation").

**Fix**: both removal paths now log a `"Remove Timed Service"` encounter-audit event (before/after: item, times, service value → retired + removedBy), via one shared method on `InwardTimedItemController` that `SurgeryBillController` calls — the same reuse pattern already established by `retireTimedServiceBill` (its own Javadoc: *"Public because the Interim Bill's own Remove button ... has to retire the same bill rather than duplicate this"*).

## Decisions made without approval (dev-issue-unattended)

- **Category**: the new "Remove Timed Service" trigger falls into the timeline's existing **"Other"** category — no keyword in `InpatientEventHistoryController#categorize()` matches it, and I didn't add one, since sibling actions "Validate Surgery" / "Revert Surgery Validation" already fall into "Other" the same way. Adding a dedicated category/keyword felt like scope beyond the reported bug.
- **Self-review fix (NPE risk)**: a local `/code-review medium` pass (per this skill's step 8a) flagged that `loggedUser.getWebUserPerson()` can be `null` — a real, already-guarded state elsewhere (`SessionController#selectDepartment`'s own `"No person"` check) — which would NPE *after* the retirement was already persisted, aborting the action with no user-visible confirmation despite the DB already being changed. Fixed by falling back to the `WebUser`'s own name when `Person` is null.
- **Self-review fix (duplication)**: the same review flagged that I'd copy-pasted the audit-logging method into both controllers instead of reusing one, contradicting the reuse pattern the adjacent code documents. Consolidated into one public method on `InwardTimedItemController`.

## Testing

Verified end-to-end on rhLocal (`coop` DB), reached entirely through the menus (never by URL):

1. Menu → Inward → Search → Admissions → searched `BHT/57821` → **Inpatient Dashboard**.
2. **Add Timed Services** → added "QA Timed Service 23510" (a pre-existing local test item; had to set its `DEPARTMENTTYPE` to `Inward` in the local DB since it had none, otherwise the ward-flow autocomplete never lists it — this is a local test-data gap, not a bug).
3. **Event History** — confirmed baseline (no timed-service event yet).
4. Went back, removed the just-added service.
5. **Event History** again — a new **"Remove Timed Service"** row (category "Other") now appears; expanding it shows the correct before/after diff (`item`, `serviceValue`, `fromTime` → `removedBy`, `retired: true`).
6. Verified in both databases: `PATIENTITEM.RETIRED=1` in `coop`, and a matching `AUDITEVENT` row (`EVENTTRIGGER='Remove Timed Service'`, correct `PATIENTENCOUNTERID`) in `rhAudit`.
7. Re-ran steps 2–6 after the self-review fixes to confirm the refactored/de-duplicated code still works and the NPE fallback path doesn't break anything — second audit row confirmed with `removedBy: "Buddhika"`.

Only the ward-flow (`inward_timed_service_consume.xhtml`) removal button was live-tested; the Interim Bill / surgery-flow removal button (`SurgeryBillController#removeTimeService`) now calls the exact same shared, already-verified audit method, so it wasn't separately walked through in the browser.

No schema change, no new privilege, no mock data.

## Documentation

Checked `../hmis.wiki/Inpatient-Event-History.md` — no update needed. It already describes action rows generically ("room changes, discharges, edits ... and more") without claiming Timed Service removal specifically was or wasn't tracked, so nothing there was factually wrong to correct.

https://claude.ai/code/session_01LxbTvRS6L1WeMSjNQapHBT


buddhika75:
@coderabbitai review

Issue #23924, opened 2026-09-20
Custom Bill 3 (Letterhead): always print patient address, NIC, phone, guardian and corporate sponsor - remove the per-field config


buddhika75:
## Summary

On the Custom Bills tab of the inward final bill reprint page, the "Custom Bill 3 (Letterhead)"
format hides patient address, NIC, phone, guardian and corporate sponsor until the user opens the
**Settings** dialog once, after which they appear for the rest of the session.

These details should simply always print. The per-field configuration should be removed entirely.

## Environment

- Instance: QA, app version `V.3.0.0.20240813.1`; also observed on a staging instance
- Department: Inward; role: inward billing/cashier
- Code reviewed against `origin/development` @ `4232951596`

## Steps to reproduce

1. Log in fresh (new session)
2. Menu → Inward → Search → Final Bill → open any final bill → **Custom Bills** tab
3. Click **Custom Bill 3** and look at the patient detail block — address, NIC, phone, guardian and
   corporate sponsor are all absent
4. Click **Settings**, then close the dialog without changing anything
5. Click **Custom Bill 3** again — the fields now appear, and keep appearing for the rest of the
   session

## Expected

All five details print on Custom Bill 3, always, with no configuration involved.

## Actual

All five are hidden until the Settings dialog has been opened once. The behaviour looks to staff like
session-scoped configuration.

## Root cause

`BhtSummeryController` is `@SessionScoped` (`bean/inward/BhtSummeryController.java:119`). The config
loader:

```java
public void loadCustom4Config() {                                   // :740
    custom4ShowAddress   = configOptionController.getBooleanValueByKey("... Show Patient Address", false);
    custom4ShowNic       = configOptionController.getBooleanValueByKey("... Show Patient NIC", false);
    custom4ShowPhone     = configOptionController.getBooleanValueByKey("... Show Patient Phone", false);
    custom4ShowGuardian  = configOptionController.getBooleanValueByKey("... Show Guardian", true);
    custom4ShowCorporateSponsor = configOptionController.getBooleanValueByKey("... Show Corporate Sponsor", true);
}
```

has exactly one UI call site — the dialog's own open event:

```xhtml
<!-- inward/inward_reprint_bill_final.xhtml:857 -->
<p:ajax event="open" listener="#{bhtSummeryController.loadCustom2Config}" update="custom2ConfigForm" />
```

(`loadCustom2Config()` chains to `loadCustom4Config()` at `:537`.) The only other callers are the
[...]

PR #23936, opened 2026-09-21
fix(inward): list every credit company on Custom Bill 3 (Letterhead)

### LINKED ISSUE #23925: Custom Bill 3 (Letterhead): prints only one credit company when the admission has several
## Summary

The "Custom Bill 3 (Letterhead)" inward final bill format prints only one credit company, even when
the admission has several registered. The template reads the single scalar `bill.creditCompany`
instead of iterating the admission's `EncounterCreditCompany` collection.

## Environment

- Instance: QA (`qa2`), app version `V.3.0.0.20240813.1`
- Department: Inward; role: inward billing/cashier
- Code reviewed against `origin/development` @ `4232951596`

## Steps to reproduce

1. Use an admission that has **two** credit companies registered against it
2. Menu → Inward → Search → Final Bill → open that admission's final bill → **Custom Bills** tab
3. Click **Custom Bill 3**
4. Read the "Credit Company" line in the patient detail block

## Expected

All credit companies registered against the admission are listed on the bill.

## Actual

A single credit company is printed. The same single value repeats across every bill copy
(Original, Duplicate, …). The sponsor totals block likewise shows one combined set of figures —
Total Sponsor Amount / Sponsor Discount Amount / Sponsor Amount — with no per-company breakdown.

## Root cause

`src/main/webapp/resources/inward/bill/finalBillCustom4.xhtml:175-183` renders a scalar, not a list:

```xhtml
<h:panelGroup layout="block" styleClass="col-6 d-flex mb-1"
              rendered="#{bhtSummeryController.custom4ShowCorporateSponsor
                          and cc.attrs.bill.creditCompany ne null
                          and cc.attrs.bill.paymentMethod eq 'Credit'}">
    <span style="width: 130px;">Credit Company</span>
    <span style="width: 12px; text-align: center;">:</span>
    <span class="flex-fill ps-2">
        <h:outputLabel value="#{cc.attrs.bill.creditCompany.name}" />
    </span>
</h:panelGroup>
```

`Bill.creditCompany` is a single `Institution`, as is `PatientEncounter.creditCompany`
(`core/entity/PatientEncounter.java:94`). The multi-sponsor data lives in a **separate collection** —
the `EncounterCreditCompany` entity (`core/entity/EncounterCreditCompany.java`) — which the billing
code already iterates when it settles each sponsor:
`BhtSummeryController.saveCCBill(PatientEncounter, EncounterCreditCompany, Double)` at
`bean/inward/BhtSummeryController.java:3965`, driven from the list returned by
`fillCreditCompaniesByPatient(patientEncounter)` (see `:3012`).

So the model supports many sponsors and the settlement path honours them; only this print template
assumes one.

## Fix direction

Iterate `fillCreditCompaniesByPatient(bill.patientEncounter)` in the template rather than reading
`bill.creditCompany`.

Two points to settle before implementing:

1. **Should the sponsor amount totals also split per company?** With two sponsors a single combined
   "Sponsor Amount" is ambiguous. Not requested explicitly, so confirm before changing it.
2. The render condition also requires `bill.paymentMethod eq 'Credit'`. An admission with sponsors
   but a non-Credit payment method on the bill would show nothing at all — worth confirming whether
   that gate is still correct once the list is rendered.

## Related

Same file and same detail block as the Custom Bill 3 configuration-loading defect filed separately;
the "Show Corporate Sponsor" toggle is what reveals this block.

buddhika75:
Closes #23925

> **Stacked on #23935** (issue #23924). Both edit `finalBillCustom4.xhtml`; this branch is cut from that one, so please merge #23935 first. The diff against `development` will include #23935's commit until then.

## Problem

Custom Bill 3 printed **one** credit company even when the admission had several registered, and the same single value repeated across every bill copy. The template read the scalar `bill.creditCompany`; the multi-sponsor data lives in the separate `EncounterCreditCompany` collection, which the settlement path (`saveCCBill` / `fillCreditCompaniesByPatient`) already iterates. The model supports many sponsors — only this print template assumed one.

## Change

**`BhtSummeryController.creditCompanyNamesForPrint(Bill)`** (new) — reads the admission's `EncounterCreditCompany` rows and returns their names as a comma-separated string.

- **De-duplicates.** Duplicate registrations of the same institution against one admission are possible through older data paths — `rebuildAllocationsFromSourceBill` already carries a comment about exactly this — so a company registered twice is listed once rather than printed twice.
- **Falls back to the scalar.** When the admission has no `EncounterCreditCompany` rows at all, the method returns `bill.creditCompany`'s name, so older admissions keep printing the sponsor they always did.
- Returns `""` when there is no sponsor, which the template uses to hide the whole line.

**`finalBillCustom4.xhtml`:**

- The Credit Company line renders that string instead of `bill.creditCompany.name`.
- **The `bill.paymentMethod eq 'Credit'` gate is removed.** The sponsors are a fact of the admission; the payment method recorded on the bill is a separate thing, and gating on it meant an admission with registered sponsors could print none at all. The line now renders whenever there is a sponsor to show, and stays hidden when there is not — so a bill with no sponsor is unchanged.
- Each line of the **Paid By Company** block now carries the paying company's name next to the bill number and amount, so the per-company figures already printed there become attributable.

### A note on the issue's description

The issue describes a "Total Sponsor Amount / Sponsor Discount Amount / Sponsor Amount" block on Custom Bill 3. Those labels are **not** in `finalBillCustom4.xhtml` — they belong to `finalBillCustom2.xhtml` (Custom Bill 2). What Custom Bill 3 actually has is the Paid By Company block described above, so that is what was made per-company. No new totals section was invented on this format.

## Verification (local Payara, local DB)

Test data was built through the app, not SQL. On an existing final bill's admission I used **Create New Version → Add Credit Company** to register two *different* sponsors (Softlogic Insurance, Ceylinco Life) — confirmed as two `EncounterCreditCompany` rows in the DB.

Then: **Menu → Final Bill Search → Search Bills → click the bill number → Custom Bills tab → Custom Bill 3**.

The rendered Credit Company line, on every bill copy:

```
Credit Company : Softlogic Insurance, Ceylinco Life
```

Both sponsors, comma-separated, where the old template rendered a single scalar.

Honest scope of what was exercised live:

- **Multiple sponsors listed** — verified, above.
- **De-duplication** — covered by the `LinkedHashSet` in the helper; not separately rendered, as the two test sponsors are distinct.
- **Dropped payment-method gate** — not separately rendered: it needs a bill that is *not* Credit on an admission that *has* sponsors, which no local record is in. The behaviour follows directly from the removed condition.

No screenshot is published: the local DB is a production-derived copy, so the rendered bill shows a real patient's name, address and phone. Test data was left in place.

## Documentation

[Finance — Inpatient Final Bill](https://github.com/hmislk/hmis/wiki/Finance-Inpatient-Final-Bill) — the Custom Bill 3 section now states that all sponsors are listed, that the line does not depend on the bill's payment method, and that a company registered twice is listed once.

## Summary by CodeRabbit

* **New Features**
  * Bills now display all available patient details, including address, NIC, phone number, and guardian information.
  * Credit-company information can display multiple sponsors in registration order.
  * Sponsor-paid bill lines now identify the corresponding company.

* **Changes**
  * Custom Bill 3 letterhead display settings are no longer available in the printer configuration dialog.


buddhika75:
Applied the ordering fix — thanks.

`fillCreditCompaniesByPatient()` now ends with `order by ecc.id`, so the sponsors come back in registration order and that order reaches the printed bill deterministically through `creditCompanyNamesForPrint()`.

One deviation from the suggestion: ordered by `id` alone rather than `createdAt ASC, id ASC`. `id` is monotonic and never null, so it *is* registration order; `createdAt` is nullable on older rows, and JPQL/MySQL would sort those NULLs to the front, which would put legacy registrations ahead of newer ones — the opposite of what the finding asks for.

